PR #2, opened 2017-09-04
2.0.0 (Initial) Release

nblumhardt:
Initial version of the package (version to avoid any confusion as to which version of ASP.NET Core is supported).


nblumhardt:
It would be nice to get #1 in for this, however it is an additive change and can be shipped in a 2.1.x.

Issue #5, opened 2017-09-25
Using appsettings to configure


nrandell:
Not really an issue, but I've just experimented with using appsettings to configure the logger. I don't know if this is something worth adding in or not, so I'll just include a few code snippets. 

Happy to create a PR if it's valuable.

First off, in the extensions
```
public static IWebHostBuilder UseSerilogLogger(this IWebHostBuilder builder)
{
    if (builder == null) throw new ArgumentNullException(nameof(builder));
    builder.ConfigureServices((context, collection) =>
        collection.AddSingleton<ILoggerFactory>(new LoggerFactory(context.Configuration))
    );
    return builder;
}
```

Then in the factory
```
public LoggerFactory(IConfiguration configuration)
{
    Log.Logger = new LoggerConfiguration()
        .ReadFrom.Configuration(configuration)
        .CreateLogger();
    _provider = new SerilogLoggerProvider();
}
```

And finally my settings
```
  "Serilog": {
    "MinimumLevel": {
      "Default": "Debug",
      "Override": {
        "Microsoft": "Warning",
        "System": "Warning"
      }
    },
    "WriteTo": [
      {
        "Name": "Console"
      }
    ],
    "Enrich": [ "FromLogContext", "WithMachineName", "WithThreadId" ]
  }
```

cyrusjamula:
This would be great

nrandell:
Ok - working on it

nblumhardt:
Thanks for the suggestion and discussion 👍 

I am a bit hesitant to start wrapping Serilog via this package; the Serilog API already provides for this case, e.g. check out this sample: https://github.com/serilog/serilog-aspnetcore/blob/dev/samples/SimpleWebSample/Program.cs#L13

There are some rough edges, discussed in #1, but in general I think we're aiming to wrap as little of Serilog as possible, just to get the logging provider replaced.

nrandell:
I'd seen that one - the thing that made me do the initial work was to keep the `WebHost.CreateDefaultBuilder(args)` without having to manually create the configuration. Maybe what's needed is an extension method that takes a configuration object and returns the logger as one of the parameters. 

eg. `public static IWebHostBuilder UseSerilog(this IWebHostBuilder builder, Func<IConfiguration, Serilog.ILogger> logBuilder, bool setCoreLogger = true, bool dispose = false)`

That way there are no more dependencies added, but it keeps everything simple. Does that make sense?

nrandell:
[...]

PR #7, opened 2017-09-27
Added support for appsettings configuration

nrandell:
PR for issue #5 that adds in functionality to use the appsettings in a single line

### DIFF
--- a/samples/ConfigurationWebSample/ConfigurationWebSample.csproj
+++ b/samples/ConfigurationWebSample/ConfigurationWebSample.csproj
@@ -1,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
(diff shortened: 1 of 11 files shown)


nrandell:
I've removed the additional dependency to `serilog.settings.configuration`, and created a logger builder extension to add serilog. This keeps the original package nice and clean.
The example now uses
```
.UseSerilog(configuration =>
    new Serilog.LoggerConfiguration()
        .ReadFrom.Configuration(configuration)
        .CreateLogger()
)
```

nblumhardt:
Thanks for pushing ahead with this one, @nrandell 👍 

I think this has become effectively a duplicate of #1, where there's already been a bit more discussion around trade-offs. I think we should close this PR (sorry!) and pick things up over there. Sound like a plan?

nrandell:
I've just read #1 a bit more carefully @nblumhardt and it makes sense. So I'm happy to close this - it's good to see that a number of us have come to the same conclusions separately! We must be doing something right!

PR #12, opened 2017-10-21
2.1.0 Release

nblumhardt:
* #1 - option to configure inline so that default configuration/hosting context information is available

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Then, add `UseSerilog()` to the web host builder. A `try`/`catch` block will ens
 **Finally**, clean up by removing the remaining configuration for the default logger:
 
  * Remove calls to `AddLogging()`
- * Remove the `"Logging"` section from _appsettings.json_ files (this can be replaced with [Serilog configuration](https://github.com/serilog/serilog-settings-configuration) as shown in [this example](), if required)
+ * Remove the `"Logging"` section from _appsettings.json_ files (this can be replaced with [Serilog configuration](https://github.com/serilog/serilog-settings-configuration) as shown in [this example](https://github.com/serilog/serilog-aspnetcore/blob/dev/samples/SimpleWebSample/Program.cs), if required)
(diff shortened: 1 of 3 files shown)


toddlucas:
This looks like a step in the right direction. I would like to set up logging in `Startup`. A big part of what this method is doing is adding the factory singleton to services, which often happens inside `Startup.ConfigureServices`. In Core 2.0 the pattern seems to be:

```
ConfigureServices(IServiceCollection services) { 
    services.AddMyService();
    services.AddMvc();
} 
```

and in `Startup.Configure`:

```
Configure(IApplicationBuilder app) {
    app.UseMyService();
    app.UseMvc(); 
}
```

If the `UseSerilog()` extension wrapped a new `AddSerilog()` extension which would add the factory, then it would be possible to call `services.AddSerilog()` from `ConfigureServices` as above. 

I would write my own, but right now I can't create an extension like that because `SerilogLoggerFactory` isn't public. I can write my own factory, since `SerilogLoggerProvider` is available, but it would be nice to use the one from this project. Adding the `AddSerilog` extension would make it possible to follow the new typical pattern. Thanks!

nblumhardt:
Hi @toddlucas - we're keen to keep the top-level `UseSerilog()` pattern for now, but thanks for the suggestion - will keep it in mind. Cheers! 👍

toddlucas:
Thanks @nblumhardt. If you're amenable, I might create a PR to show you an example. The `UseSerilog()` pattern wouldn't need to change.

nblumhardt:
Thanks for the follow-up, Todd. Just to zoom out for a second, what't the advantage to be gained by configuring Serilog in Startup.cs? Thanks!

toddlucas:
Actually, I've taken another look and the Add/Use way appears to be more of a 1.x convention--at least for logging. 2.x has moved default configuration and logging setup to Program.cs. It's often hidden in templates that use `CreateDefaultBuilder()`. It makes sense to be able to set it up earlier, before Startup is invoked. Please disregard.

nblumhardt:
Okay, thanks for the info 👍

PR #19, opened 2017-12-09
Expose `SerilogLoggerFactory`

jonathansant:
I needed to retrieve some configuration prior to registering the logger. `UseSerilog` is invoked before the `Startup` class. I sought to register the Factory manually but found out that is internal. I think that `SerilogLoggerFactory` is the main take away from this library and therefore exposing it would make it more flexible for those who don't want to use the extension methods.

### DIFF
--- a/src/Serilog.AspNetCore/AspNetCore/SerilogLoggerFactory.cs
+++ b/src/Serilog.AspNetCore/AspNetCore/SerilogLoggerFactory.cs
@@ -18,26 +18,49 @@ using Serilog.Extensions.Logging;
 
 namespace Serilog.AspNetCore
 {
-    public class SerilogLoggerFactory : ILoggerFactory
+	/// <summary>
+	/// Implements Microsoft's ILoggerFactory so that we can inject Serilog Logger.
+	/// </summary>
+	/// <seealso cref="Microsoft.Extensions.Logging.ILoggerFactory" />
+	public class SerilogLoggerFactory : ILoggerFactory
     {
         readonly SerilogLoggerProvider _provider;
 
-        public SerilogLoggerFactory(Serilog.ILogger logger = null, bool dispose = false)
+		/// <summary>
+		/// Initializes a new instance of the <see cref="SerilogLoggerFactory"/> class.
+		/// </summary>
+		/// <param name="logger">The logger.</param>
+		/// <param name="dispose">if set to <c>true</c> [dispose].</param>
+		public SerilogLoggerFactory(Serilog.ILogger logger = null, bool dispose = false)
         {
             _provider = new SerilogLoggerProvider(logger, dispose);
         }
 
-        public void Dispose()
+		/// <summary>
+		/// Disposes the provider.
+		/// </summary>
+		public void Dispose()
         {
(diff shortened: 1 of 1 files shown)


nblumhardt:
Thanks for the PR, @jonathansant.

Does the alternative overload of `UseSerilog()`, [which accepts a delegate](#1), cover your scenario?

Best regards,
Nick

jonathansant:
Since all my startup logic is in the `Startup` class, I can ugly hack it using statics but it would be ideal if I kept all the startup logic in one place. What would help is an extension method for the `IServiceCollection` class and then I could put everything in the `ConfigureServices` method. I can create a pull request for that if you prefer.

nblumhardt:
Thanks for the reply. I think providing the factory type is fine 👍

src/Serilog.AspNetCore/AspNetCore/SerilogLoggerFactory.cs
```diff
@@ -18,26 +18,49 @@
 
 namespace Serilog.AspNetCore
 {
-    class SerilogLoggerFactory : ILoggerFactory
+	/// <summary>
+	/// Implements Microsoft's ILoggerFactory so that we can inject Serilog Logger.
+	/// </summary>
+	/// <seealso cref="Microsoft.Extensions.Logging.ILoggerFactory" />
+	public class SerilogLoggerFactory : ILoggerFactory
```
nblumhardt: Indentation is funny here and below - may need to use spaces, rather than tabs?

src/Serilog.AspNetCore/AspNetCore/SerilogLoggerFactory.cs
```diff
@@ -18,26 +18,49 @@
 
 namespace Serilog.AspNetCore
 {
-    class SerilogLoggerFactory : ILoggerFactory
+	/// <summary>
+	/// Implements Microsoft's ILoggerFactory so that we can inject Serilog Logger.
```
nblumhardt: Instead of `Microsoft's ILoggerFactory` we could use `<see cref="ILoggerFactory"/>`

src/Serilog.AspNetCore/AspNetCore/SerilogLoggerFactory.cs
```diff
@@ -18,26 +18,49 @@
...
+	public class SerilogLoggerFactory : ILoggerFactory
     {
         readonly SerilogLoggerProvider _provider;
 
-        public SerilogLoggerFactory(Serilog.ILogger logger = null, bool dispose = false)
+		/// <summary>
+		/// Initializes a new instance of the <see cref="SerilogLoggerFactory"/> class.
+		/// </summary>
+		/// <param name="logger">The logger.</param>
+		/// <param name="dispose">if set to <c>true</c> [dispose].</param>
```
nblumhardt: Could use the same documentation here as in https://github.com/serilog/serilog-aspnetcore/blob/dev/src/Serilog.AspNetCore/SerilogWebHostBuilderExtensions.cs#L33

nblumhardt:
Thank you 👍

PR #56, opened 2018-07-18
Added information about the Serilog.Settings.Configuration package

patriksvensson:
`Serilog.Settings.Configuration` needs to be installed for the delegate configuration example to work.


nblumhardt:
Thanks 👍

PR #81, opened 2019-03-06
Update sample to ASP.NET Core 2.2

nblumhardt:


nblumhardt:
Closing, we should delete/re-create an idiomatic sample instead.

PR #87, opened 2019-05-22
Integrate with LoggerProviderCollection

### LINKED ISSUE #53: Move SerilogLoggerFactory to Serilog.Extensions.Logging
From a distance, it seems that while `SerilogWebHostBuilderExtensions` deals with bona fide ASP.NET concerns (e.g `IWebHostBuilder`), `SerilogLoggerFactory` would be better placed in a more generic .NET Core library.  Perhaps `Serilog.Extensions.Logging`.

There's also the possibility that I'm completely insane. :)

Of course if not, I realize this would be a breaking change but maybe something to consider in the next major version of this library.

nblumhardt:
This will integrate `UseSerilog()` with the `ILoggerProvider` support in upstream _Serilog.Extensions.Logging_.

For the more self-contained (post-IoC-initialization) variant, it'll be as simple as `UseSerilog(writeToProviders: true)`.

When logging is configured first, at program startup, a `LoggerProviderCollection` will need to be constructed and passed to `WriteTo.Providers()` as well as `UseSerilog()`.

For some context, this will enable usage like: https://github.com/opentracing-contrib/csharp-netcore/issues/42

Resolves #53

@andrewlock I'm hoping this will translate cleanly across to _Serilog.Extensions.Hosting_, and once we're happy with it here, I'll assemble a second PR :-)

 - [x] Working
 - [ ] ~~Tests~~ Turns out there isn't really any (sanely) testable implementation in this package, it's all now in _Serilog.Extensions.Logging_
 - [ ] ~~_Should the second `UseSerilog()` overload default to writing to providers?_~~ No :-)
 - [x] README updates
 - [x] Make sure providers added after logger initialization are still added to the LPC


nblumhardt:
Resolves #53

nblumhardt:
I know it's currently tricky for everyone to keep abreast of these changes; I think as in _Serilog.Extensions.Logging_ our best bet is to send them through to `dev` and feel our way forwards from there.

Just one thing to call out - this changes the signature of `UseSerilog()` in a source, but not binary-compatible way. I think this is acceptable for this package, on a major version update, as it's unlikely to be involved in any nasty diamond dependency scenarios.

I'll hit the button in a day or so unless I hear from anyone.

PR #90, opened 2019-06-04
Update to initial configuration pattern

pmorong:
Update to the initial configuration pattern so that it matches up-to-date asp.net core pattern.

ref: https://docs.microsoft.com/en-us/aspnet/core/migration/20_21?view=aspnetcore-2.1#changes-to-main


nblumhardt:
Thank you 👍

PR #96, opened 2019-06-26
README and dependency updates

nblumhardt:
Updated to use the latest _Serilog.Extensions.Hosting_; describe request logging in the README.


nblumhardt:
Still in churn mode while getting the fundamentals of this feature in order. I'll send it through the works and follow up on any feedback afterwards 😊

Issue #95, opened 2019-06-25
README and sample updates for v3.0


nblumhardt:
Once v3.0 of this project ships, we should show use of `IHostBuilder` through _Serilog.Extensions.Hosting_ instead of the `IWebHostBuilder` code we continue to support here for backward compatibility. There are also some new features to document.

 - [ ] Update the README and samples here to use `IHostBuilder` and current ASP.NET Core conventions
 - [ ] Update the _Serilog.Extensions.Hosting_ README to replicate the more detailed information about early vs. inline initialization, and provider support, currently documented in this project
 - [x] Add documentation for `IDiagnosticContext`
 - [x] Include `Debug`, `Console`, and `File` sinks as dependencies, with example usage in the README; `Settings.Configuration` may also be worth consideration

carlowahlstedt:
Not sure if this is for compatibility with Core 3, but enabling Serilog is eating the output to the debug console after upgrading. Here's my Program.cs post upgrade where I had to comment out Serilog to get the output to the debug console to work.

![image](https://user-images.githubusercontent.com/2687687/63166958-d7395280-bffd-11e9-9763-1c887482ddc5.png)

I couldn't find any other way to enable Serilog with the extension method, docs, or anything else that gave be an idea of what I might be doing wrong.

nblumhardt:
Thanks for the heads-up. You need `WriteTo.Debug()` in your Serilog configuration if you want log events to go to the debug console (`dotnet add package Serilog.Sinks.Debug`) - it wouldn't hurt if we take a dependency on this and the console sink from this package, I'll make that change and include this info in the README. Cheers!

carlowahlstedt:
Thanks for the quick response. I only noticed because opening of the browser window in vs code stopped working after the update. Finally tracked it back to that.

rogerfar:
Is there an upgrade path needed from 2.x to 3.0? It looks like the CHANGES.MD isn't kept up to date.

nblumhardt:
Hi @rogerfar - thanks for pointing out the unmaintained `CHANGES.md`. I've just deleted it - changes are now tracked per release in https://github.com/serilog/serilog-aspnetcore/releases

There are no API-breaking changes in v3.0, though some functionality has moved around between packages under the hood.

If you encounter any issues upgrading, let us know, we'll pop some notes into the 3.0.0 milestone release notes. Thanks!

nblumhardt:
#111 is tracking the last of the example changes.

PR #110, opened 2019-08-16
Dependency and README updates

nblumhardt:
This PR adds `Console`, `Debug`, and `File` sinks to the package dependencies, along with _Serilog.Formatting.Compact_ (JSON) and _Serilog.Settings.Configuration_. The README is updated accordingly.

The goal is to make the onboarding process smoother. There's still work to do, including sample updates to match the ASP.NET Core 3.0 hosting startup.

Partial #95


samples/EarlyInitializationSample/Program.cs
```diff
@@ -21,6 +21,7 @@ public static int Main(string[] args)
             Log.Logger = new LoggerConfiguration()
                 .ReadFrom.Configuration(Configuration)
                 .Enrich.FromLogContext()
+                .WriteTo.Debug()
```
merbla: I like this, there are lots of questions when folks are getting started that this may help.

nblumhardt: :+1:

nblumhardt:
The reason I left this in draft is that it will update the README in a way that's incompatible with the current released package. We could just jump ahead, however, and release our 3.0 with the framework dependencies left at 2.2 (it doesn't look like we'll bump dependency versions this time around). Thoughts?

SimonCropp:
> We could just jump ahead

:shipit:

nblumhardt:
Will just need: https://github.com/serilog/serilog-extensions-hosting/pull/11

nblumhardt:
Planning to merge and release this via 3.0.0, and only merge the example updates in #111 once .NET Core 3 has shipped.

PR #112, opened 2019-08-20
update to stable Serilog.Extensions.Logging v3

SimonCropp:


nblumhardt:
👍 

Since this package depends on _Serilog.Extensions.Hosting_, which itself depends on _Logging_, should we still have this dependency?

When https://github.com/serilog/serilog-extensions-hosting/pull/13 is through, updating to the 3.0.0 of _Hosting_ will pull the correct version of _Logging_ through for us. What do you think?

SimonCropp:
@nblumhardt happy with that. want a new PR that removes Extensions.Logging ?

nblumhardt:
That sounds great 👍

PR #113, opened 2019-08-20
update to 2.2.401 sdk

SimonCropp:


nblumhardt:
#111 takes us a step further and uses the latest 3.0 preview SDK, not sure how we want to map all of this out, but worth considering that we'll have another change coming in the near future.

SimonCropp:
@nblumhardt TBH i woul prefer to remove that setting. it make it extremely difficult to contribute to OSS when projects are hard coded to a specific version of the SDK.

nblumhardt:
Agreed it can be a pain. Using an unpinned SDK version was painful in other ways (inconsistent builds, breakage), it's a shame there's no "best effort" SDK matching in IDEs 🤔

SimonCropp:
but given #111 this makes no sense

PR #116, opened 2019-08-20
update test and sample refs

SimonCropp:
i also smoke tested


nblumhardt:
Great 👍 - also some overlap with #111 but no reason to hold back :-)

Issue #123, opened 2019-08-28
RequestLoggingMiddleware does not add a SourceContext when logging


adamlith:
This makes it unusable for us as we filter logmessages based on source context and log level.

nblumhardt:
Thanks for the note; I'd considered this originally but thought I'd lean towards smaller event payloads without it, but in hindsight it should probably be there 👍

alexvaluyskiy:
@nblumhardt have you plan to release this bugfix soon? 
Without `Log.ForContext<RequestLoggingMiddleware>()` we can't override a log level by a namespace

nblumhardt:
Hi @alexvaluyskiy - have you tried the current pre-release (`-dev-*`) package, and does it work as required for you? I'll open a release PR now, keen to have any feedback/validation of the prerelease that you can send.

alexvaluyskiy:
@nblumhardt yes, the current `dev` version works great

PR #124, opened 2019-08-29
Adding source context for the logging in RequestLoggingMiddleware. 

adamlith:
Fixing Issue #123

### DIFF
--- a/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
+++ b/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
@@ -70,9 +70,10 @@ namespace Serilog.AspNetCore
 
         bool LogCompletion(HttpContext httpContext, DiagnosticContextCollector collector, int statusCode, double elapsedMs, Exception ex)
         {
+            var logger = Log.ForContext<RequestLoggingMiddleware>();
             var level = statusCode > 499 ? LogEventLevel.Error : LogEventLevel.Information;
 
-            if (!Log.IsEnabled(level)) return false;
+            if (!logger.IsEnabled(level)) return false;
             
             if (!collector.TryComplete(out var collectedProperties))
                 collectedProperties = NoProperties;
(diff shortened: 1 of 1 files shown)


src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
```diff
@@ -70,9 +70,10 @@ public async Task Invoke(HttpContext httpContext)
 
         bool LogCompletion(HttpContext httpContext, DiagnosticContextCollector collector, int statusCode, double elapsedMs, Exception ex)
         {
+            var logger = Log.Logger.ForContext<RequestLoggingMiddleware>();
```
nblumhardt: This will cause some additional allocations on every request; alternatively, we could create a:

```csharp
static readonly LogEventProperty SourceContextProperty = new LogEventProperty("SourceContext", typeof(SerilogRequestLoggingMiddleware).FullName);
```

and then include this in the list of properties that are manually added to the event a few lines below?

adamlith: But do we not need to have set the source context for the IsEnabled chack?
```csharp
if (!logger.IsEnabled(level)) return false;
```
 if not then I agree with you and I'll change it to your suggestion.

nblumhardt: Ah right you are, nicely spotted 😅

Would still be nice to trim some allocations down the track, but I think the change as written is good for now 👍 

Very minor nitpick:

```
var logger = Log.Logger.ForContext<RequestLoggingMiddleware>();
```

It's usually idiomatically written as:

```
var logger = Log.ForContext<RequestLoggingMiddleware>();
```

adamlith: Alright, I've fixed the nitpick-issue 👍

nblumhardt:
👍 great, thanks Adam!

Issue #126, opened 2019-08-31
WebHost inside Generic host not logging startup messages


gthvidsten:
If you have a generic host, then inside this host you add a web host, and you are using `.UseSerilog()` for both, then the web host startup messages are not logged to Serilog.

This is what I mean by "startup messages":
> Hosting environment: Production
> Content root path: C:\Users\Me\Project\src\App\bin\Debug\netcoreapp3.0
> Now listening on: http://localhost:5000
> Application started. Press Ctrl+C to shut down.

**Program.cs**
```csharp
public static async Task Main(string[] args)
{
    Log.Logger = new LoggerConfiguration()
        .MinimumLevel.Debug()
        .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
        .Enrich.FromLogContext()
        .WriteTo.Console()
        .CreateLogger();

    IHostBuilder builder = new HostBuilder()
        .UseSerilog()
        .ConfigureServices(services => { collection.AddHostedService<MyService>(); });

    await builder.RunConsoleAsync();
}
```
**MyService.cs**
```csharp
public class MyService : IHostedService
{
    private IWebHost _webHost;

    public async Task StartAsync(CancellationToken ct)
    {
        _webHost = WebHost.CreateDefaultBuilder()
            .UseStartup<MyService>()
            .UseSerilog()
            .Build();

        _webHost.RunAsync();
    }

    public async Task StopAsync(CancellationToken ct)
    {
        await _webHost.StopAsync();
    }

    public void ConfigureServices(IServiceCollection services) {
    }

    public void Configure(IApplicationBuilder app, IWebHostEnvironment env) {
    }
}
```
When this starts the logging from the generic host uses Serilog, but the logging in the web host startup does not. Logging of requests to the web host, on the other hand, does use Serilog. This is apparent in the output:

> Hosting environment: Production
> Content root path: C:\Users\Me\Project\src\App\bin\Debug\netcoreapp3.0
> Now listening on: http://localhost:5000
> Application started. Press Ctrl+C to shut down.
> [15:32:01 INF] Application started. Press Ctrl+C to shut down.
> [15:32:01 INF] Hosting environment: Production
> [15:32:01 INF] Content root path: C:\Users\Me\Project\src\App\bin\Debug\netcoreapp3.0\
> [15:32:33 INF] Request starting HTTP/1.1 GET http://localhost:5000/
> [15:32:33 INF] Request finished in 13.2002ms 404

[...]

PR #127, opened 2019-09-02
Remove unnecessary dependencies

sungam3r:
image
I think that these dependencies were added unintentionally due to sample projects that reference the main package.

In fact, only two are enough:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ Finally, pass the provider collection into `UseSerilog()`:
 
 Providers registered in _Startup.cs_ with `AddLogging()` will then receive events from Serilog.
 
-**Using iniline initialization:**
+**Using inline initialization:**
 
 If [inline initialization](#inline-initialization) is used, providers can be enabled by adding `writeToProviders: true` to the `UseSerilog()` method call:
 
--- a/src/Serilog.AspNetCore/Serilog.AspNetCore.csproj
+++ b/src/Serilog.AspNetCore/Serilog.AspNetCore.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Description>Serilog support for ASP.NET Core logging</Description>
@@ -21,16 +21,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Serilog" Version="2.8.0" />
     <PackageReference Include="Serilog.Extensions.Hosting" Version="3.0.0" />
-    <PackageReference Include="Serilog.Sinks.Console" Version="3.1.1" />
-    <PackageReference Include="Serilog.Sinks.File" Version="4.0.0" />
-    <PackageReference Include="Serilog.Sinks.Debug" Version="1.0.1" />
-    <PackageReference Include="Serilog.Settings.Configuration" Version="3.1.0" />
-    <PackageReference Include="Serilog.Formatting.Compact" Version="1.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.0.0" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.0.0" />
   </ItemGroup>
 
 </Project>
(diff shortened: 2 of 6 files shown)


nblumhardt:
Thanks Ivan! This is #110 - moving this package towards a "batteries included" approach, since the core `UseSerilog()` functionality is now implemented in the (low dependency) _Serilog.Extensions.Hosting_ package.

It would be interesting to know which features of _Serilog.AspNetCore_ you're using, so that we can evaluate whether this is the optimal factoring. Cheers!

sungam3r:
Wow! I could not even imagine that it was intended. I highly recommend not including such dependencies. It's like shooting yourself in the foot. With the same success, you can turn on a bunch of other sinks/formatters/appconfigs/etc. in the hope that they will be useful to someone. But in fact, only version conflicts will arise when the consumer uses a package of another major version. These are all spurious dependencies.

> The goal is to make the onboarding process smoother.

In this case, you can make a separate package and name it like `Serilog.AspNetCore.All` but I still would not recommend doing so.

For some consumers, this concern actually turns into problems of versioning dependencies. And the other part simply does not use this "ballast". You have already done sample projects with good examples. No need to mix.

sungam3r:
> It would be interesting to know which features of Serilog.AspNetCore you're using, so that we can evaluate whether this is the optimal factoring.

I use `Serilog.Sinks.Kafka` for example.

nblumhardt:
Thanks for the reply!

As you'd know, the Serilog packages are all very fine-grained, for many of the reasons you point out. Unfortunately, something resembling a meta-package is needed if we're going to meet the ease-of-setup expectations set by _Microsoft.Extensions.Logging_/ASP.NET Core, which includes console and debug logging by default. 

> I use Serilog.Sinks.Kafka for example.

Sorry, I meant which features _from this package_? Since the core logging adapters and `UseSerilog()` are now in _Serilog.Extensions.Logging_ and _Serilog.Extensions.Hosting_, the only feature actually implemented directly in this package is `UseSerilogRequestLogging()` - unless you're using that middleware, then _Serilog.Extensions.Hosting_ is possibly the dependency-free package that you're looking for. Let me know! 👍 

None of this set in stone, of course! It will be interesting to see whether the batteries-included approach here reduces some of the support around this package, or ends up increasing it. We can definitely re-evaluate and change direction if needed.

src/Serilog.AspNetCore/Serilog.AspNetCore.csproj
```diff
@@ -21,16 +21,8 @@
...
-    <PackageReference Include="Serilog" Version="2.8.0" />
     <PackageReference Include="Serilog.Extensions.Hosting" Version="3.0.0" />
-    <PackageReference Include="Serilog.Sinks.Console" Version="3.1.1" />
-    <PackageReference Include="Serilog.Sinks.File" Version="4.0.0" />
-    <PackageReference Include="Serilog.Sinks.Debug" Version="1.0.1" />
-    <PackageReference Include="Serilog.Settings.Configuration" Version="3.1.0" />
-    <PackageReference Include="Serilog.Formatting.Compact" Version="1.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.0.0" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.0.0" />
```
nblumhardt: We should in any case be able to get rid of these two 👍

README.md
```diff
@@ -205,7 +205,7 @@ Finally, pass the provider collection into `UseSerilog()`:
 
 Providers registered in _Startup.cs_ with `AddLogging()` will then receive events from Serilog.
 
-**Using iniline initialization:**
+**Using inline initialization:**
```
nblumhardt: Argh, nice catch 😅

sungam3r:
> unless you're using that middleware, then Serilog.Extensions.Hosting is possibly the dependency-free package that you're looking for

When I read this comment, I realized that this is what I needed. But when I opened the source code, I was disappointed. It uses `IHostBuilder`, but I need `IWebHostBuilder`. So `Serilog.Extensions.Hosting` package only integrates logging to Generic Host and the package name can be misleading.

Relates to #126

nblumhardt:
Ah, thanks for the follow-up. This is a point-in-time issue, I think, since in .NET Core 3.0, startup will be `HostBuider`-based (with `WebHostBuilder` deprecated), so generally the `IHostBuilder` will be the desirable place to configure logging - unless I've misunderstood how these are intended to evolve - does that match your understanding?

sungam3r:
Yes, but it will be only in netcoreapp3.0 and netstandard2.1. Still need to wait for the release. For netstandard2.0 there is no such unification.

nblumhardt:
Thanks for the reply. It sounds like there's a bit of a compromise necessary for the time being on `netcoreapp2.x` 🤔

nblumhardt:
3.0 is now available, I think that's the way forwards. Cheers!

PR #130, opened 2019-09-05
New overload with IWebHostBuilder

sungam3r:
We **may** need access to `IServiceCollection`. Consider such use:
```csharp
 public class Program
    {
        public static void Main(string[] args)
        {
            CreateWebHostBuilder(args).Build().Run();
        }

        public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
            WebHost.CreateDefaultBuilder(args)
                .UseStartup<Startup>()
                .UseSerilog((builder, hostingContext, loggerConfiguration) =>
                {
                    var cfg = loggerConfiguration
                        .ReadFrom.Configuration(hostingContext.Configuration)
                        .Enrich.FromLogContext()
                        .WriteTo.Debug()
                        .WriteTo.Console(
                            // {Properties:j} added:
                            outputTemplate: "[{Timestamp:HH:mm:ss} {Level:u3}] {Message:lj} " +
                                            "{Properties:j}{NewLine}{Exception}");

                    //
                    builder.ConfigureServices(services => services.AddSingleton(cfg));
                });
    }
```

### DIFF
--- a/src/Serilog.AspNetCore/SerilogWebHostBuilderExtensions.cs
+++ b/src/Serilog.AspNetCore/SerilogWebHostBuilderExtensions.cs
@@ -39,8 +39,8 @@ namespace Serilog
         /// default, only Serilog sinks will receive events.</param>
         /// <returns>The web host builder.</returns>
         public static IWebHostBuilder UseSerilog(
-            this IWebHostBuilder builder, 
-            ILogger logger = null, 
+            this IWebHostBuilder builder,
+            ILogger logger = null,
             bool dispose = false,
             LoggerProviderCollection providers = null)
         {
@@ -89,6 +89,28 @@ namespace Serilog
             bool preserveStaticLogger = false,
             bool writeToProviders = false)
         {
+            return builder.UseSerilog((_, context, configuration) => configureLogger(context, configuration), preserveStaticLogger, writeToProviders);
+        }
+
+
+        /// <summary>Sets Serilog as the logging provider.</summary>
+        /// <remarks>
+        /// A <see cref="IWebHostBuilder"/> and <see cref="WebHostBuilderContext"/> are supplied so that configuration and hosting information can be used.
+        /// The logger will be shut down when application services are disposed.
+        /// </remarks>
+        /// <param name="builder">The web host builder to configure.</param>
+        /// <param name="configureLogger">The delegate for configuring the <see cref="LoggerConfiguration" /> that will be used to construct a <see cref="Logger" />.</param>
+        /// <param name="preserveStaticLogger">Indicates whether to preserve the value of <see cref="Log.Logger"/>.</param>
+        /// <param name="writeToProviders">By default, Serilog does not write events to <see cref="ILoggerProvider"/>s registered through
+        /// the Microsoft.Extensions.Logging API. Normally, equivalent Serilog sinks are used in place of providers. Specify
+        /// <c>true</c> to write events to all providers.</param>
+        /// <returns>The web host builder.</returns>
+        public static IWebHostBuilder UseSerilog(
+        this IWebHostBuilder builder,
+        Action<IWebHostBuilder, WebHostBuilderContext, LoggerConfiguration> configureLogger,
+        bool preserveStaticLogger = false,
+        bool writeToProviders = false)
+        {
             if (builder == null) throw new ArgumentNullException(nameof(builder));
             if (configureLogger == null) throw new ArgumentNullException(nameof(configureLogger));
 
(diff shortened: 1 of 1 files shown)


sungam3r:
Any thoughts?

nblumhardt:
Thanks for the PR; sorry about the slow response - juggling a few things this week and haven't had time to dig in properly yet, but I'll roll back around to this soon, if no one beats me to it.

nblumhardt:
In ASP.NET 3.0 apps we're going to be recommending `UseSerilog()` on `IHostBuilder` rather than at the web host level. I would guess that this approach can be ported/used there as well?

I think we may need to open a new meta-ticket about getting a clear picture of how all elements of the logging pipeline can be extended using DI, we've had some scattered discussions and a few approaches proposed; I'll write up what I know and link this PR when I've collected it all.

sungam3r:
@nblumhardt  We need this overload. Without it, we cannot configure serilog for our environment. This overload does not violate the current design. Do you have any thoughts against?

nblumhardt:
Hi @sungam3r , thanks for the nudge.

I'm only hesitant to add the overload because I haven't built a clear picture up yet of where this aspect of configuration is going.

In the meantime, though, since the `builder` object is the result of `WebHost.CreateDefaultBuilder(args)`, and the `UseSerilog()` argument closure can close over it, would the following workaround help?

```csharp
        public static IWebHostBuilder CreateWebHostBuilder(string[] args) => {
            var builder = WebHost.CreateDefaultBuilder(args);
            builder.UseStartup<Startup>()
                .UseSerilog((hostingContext, loggerConfiguration) =>
                {
                    var cfg = loggerConfiguration
                        .ReadFrom.Configuration(hostingContext.Configuration)
                        .Enrich.FromLogContext()
                        .WriteTo.Debug()
                        .WriteTo.Console(
                            // {Properties:j} added:
                            outputTemplate: "[{Timestamp:HH:mm:ss} {Level:u3}] {Message:lj} " +
                                            "{Properties:j}{NewLine}{Exception}");

                    //
                    builder.ConfigureServices(services => services.AddSingleton(cfg));
                });

            return builder;
        }
```

sungam3r:
I know about thiоs workaround. I just want to have a little cleaner api.

> I'm only hesitant to add the overload because I haven't built a clear picture up yet of where this aspect of configuration is going.

We have our own components and abstractions which depend from configuration. It is a sort of proxies to the real logging systems such as Serilog, NLog or others. We should set appropriate configuration in DI container as soon as possible.

sungam3r:
> I'm only hesitant to add the overload because I haven't built a clear picture up yet of where this aspect of configuration is going.

@nblumhardt  It is just another part of fluent style configuration, where client code can choose from overloads with different arguments using lambda methods. So no need in argument closures.

sungam3r:
@nblumhardt I returned to this PR and it turned out that the proposed change does not work :(

```
System.InvalidOperationException: 'Collection was modified; enumeration operation may not execute.'
   at System.ThrowHelper.ThrowInvalidOperationException_InvalidOperation_EnumFailedVersion()
   at System.Collections.Generic.List`1.Enumerator.MoveNext()
   at Microsoft.Extensions.Hosting.HostBuilder.CreateServiceProvider()
   at Microsoft.Extensions.Hosting.HostBuilder.Build()
   at WebHost.Program.Main(String[] args)
```

this is due to the fact that it is not allowed to have nested `builder.ConfigureServices()` calls.

From `HostBuilder`:
```c#
  private void CreateServiceProvider()
        {
            var services = new ServiceCollection();
            services.AddSingleton(_hostingEnvironment);
            services.AddSingleton(_hostBuilderContext);
            services.AddSingleton(_appConfiguration);
            services.AddSingleton<IApplicationLifetime, ApplicationLifetime>();
            services.AddSingleton<IHostLifetime, ConsoleLifetime>();
            services.AddSingleton<IHost, Host>();
            services.AddOptions();
            services.AddLogging();
            
            foreach (var configureServicesAction in _configureServicesActions)
            {
                configureServicesAction(_hostBuilderContext, services);
            }
```

`UseSerilog` adds callback into `_configureServicesActions`, then while executing that callback I call ` builder.ConfigureServices(services => services.AddSingleton(cfg));` inside it. So indeed I modify `List<Action>` while enumerating it. 

It is necessary to modify the collection of services in the `ConfigureWebHostDefaults` method itself. The working version looks like this:
```c#
 public static IHostBuilder CreateHostBuilder(string[] args)
        {
            var hostBuilder = Host.CreateDefaultBuilder(args);
            LoggerConfiguration loggerConfiguration = null;  // <--- store in local
            hostBuilder.ConfigureWebHostDefaults(webHostBuilder => webHostBuilder
                .UseStartup<Startup>()
                .UseSerilog((context, configuration) => loggerConfiguration = configuration.ReadFrom.Configuration(context.Configuration))
                .ConfigureServices(services => services.AddSingleton(loggerConfiguration)) // <--- add from local
            );
            return hostBuilder;
        }
```

Thus, there is no need for a new method and I close this PR.

PR #132, opened 2019-09-17
Ability to customize LogEventLevel used in UseSerilogRequestLogging middleware

seroche:
FEATURE: add ability to customize the LogEventLevel used by the middleware
FEATURE: add new dependency injection overload to offer more flexibility while configuring options


samples/InlineInitializationSample/Startup.cs
```diff
@@ -46,7 +48,14 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
             // level in appsettings.json to "Information".
-            app.UseSerilogRequestLogging();
+            app.UseSerilogRequestLogging(opts =>
+            {
+                opts.GetLogEventLevel = statusCode =>
+                    statusCode == HttpStatusCode.InternalServerError
+                        ? LogEventLevel.Error
+                        : LogEventLevel.Information;
+                // OR opts.GetLogEventLevel = _ => LogEventLevel.Information;
+            });
```
nblumhardt: I think it would still be nice to show the default, opinionated `UseSerilogRequestLogging()` in the examples, as they'll be used as templates and in most cases the defaults will be appropriate.

src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
```diff
@@ -29,15 +30,17 @@ class RequestLoggingMiddleware
         readonly RequestDelegate _next;
         readonly DiagnosticContext _diagnosticContext;
         readonly MessageTemplate _messageTemplate;
+        readonly Func<HttpStatusCode, LogEventLevel> _getLogEventLevel;
```
nblumhardt: The level might be determined from more than just the status code, would accepting `HttpContext` be more appropriate, here?

src/Serilog.AspNetCore/RequestLoggingOptions.cs
```diff
@@ -12,17 +12,37 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Serilog.Events;
 using System;
+using System.Net;
 
 namespace Serilog
```
nblumhardt: We can move this to `Serilog.AspNetCore` to avoid polluting a commonly-imported namespace; the type name doesn't have to be in scope in order for consuming code to call `UseSerilogRequestLogging(opts =>..`.

src/Serilog.AspNetCore/RequestLoggingOptions.cs
```diff
@@ -12,17 +12,37 @@
...
-            MessageTemplate = messageTemplate ?? throw new ArgumentNullException(nameof(messageTemplate));
-        }
+        /// <summary>
+        /// Gets or sets the function returning the LogEventLevel based on the HttpStatusCode.
+        /// The default behavior returns LogEventLevel.Error when HttpStatusCode is greater than 499
+        /// </summary>
+        /// <value>
+        /// The function returning the LogEventLevel.
+        /// </value>
+        public Func<HttpStatusCode, LogEventLevel> GetLogEventLevel { get; set; }
```
nblumhardt: Would `GetLevel` be an acceptable name? Elsewhere in the Serilog API, `LogEventLevel`s that appear in property/method names (`MinimumLevel`, `LogEvent.Level` etc.) are generally shortened just to `Level`.

src/Serilog.AspNetCore/SerilogApplicationBuilderExtensions.cs
```diff
@@ -26,6 +29,9 @@ public static class SerilogApplicationBuilderExtensions
         const string DefaultRequestCompletionMessageTemplate =
             "HTTP {RequestMethod} {RequestPath} responded {StatusCode} in {Elapsed:0.0000} ms";
 
+        static readonly Func<HttpStatusCode, LogEventLevel> DefaultGetLogEventLevel = 
+            s => (int) s > 499 ? LogEventLevel.Error : LogEventLevel.Information;
```
nblumhardt: This could just be a static method.

src/Serilog.AspNetCore/SerilogApplicationBuilderExtensions.cs
```diff
@@ -43,11 +49,38 @@ public static class SerilogApplicationBuilderExtensions
...
+
+        /// <summary>
+        /// Adds middleware for streamlined request logging. Instead of writing HTTP request information
+        /// like method, path, timing, status code and exception details
+        /// in several events, this middleware collects information during the request (including from
+        /// <see cref="IDiagnosticContext"/>), and writes a single event at request completion. Add this
+        /// in <c>Startup.cs</c> before any handlers whose activities should be logged.
+        /// </summary>
+        /// <param name="app">The application builder.</param>
+        /// <param name="configureOptions"> An System.Action`1 to configure the provided <see cref="Serilog.RequestLoggingOptions" />.</param>
```
nblumhardt: `A <see cref="System.Action{T}` />...`

src/Serilog.AspNetCore/RequestLoggingOptions.cs
```diff
@@ -12,17 +12,37 @@
...
-        {
-            MessageTemplate = messageTemplate ?? throw new ArgumentNullException(nameof(messageTemplate));
-        }
+        /// <summary>
+        /// Gets or sets the function returning the LogEventLevel based on the HttpStatusCode.
+        /// The default behavior returns LogEventLevel.Error when HttpStatusCode is greater than 499
+        /// </summary>
+        /// <value>
+        /// The function returning the LogEventLevel.
+        /// </value>
```
nblumhardt: Would be nice to use `<see cref=""` in the docs here to refer to other types/members (refactorings will then update docs properly/compiler will catch mistakes).

src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
seroche: I was trying to mimic the current behavior as close as possible  :). Sure no problem, I will pass the `HttpContext` to the `Func`.

samples/InlineInitializationSample/Startup.cs
seroche: Sure. no problem.

src/Serilog.AspNetCore/RequestLoggingOptions.cs
seroche: Moved the options under `Serilog.AspNetCore`. Not sure to understand what you mean by 

> the type name doesn't have to be in scope in order for consuming code to call UseSerilogRequestLogging(opts =>...

src/Serilog.AspNetCore/RequestLoggingOptions.cs
seroche: No problem.

src/Serilog.AspNetCore/SerilogApplicationBuilderExtensions.cs
seroche: My Resharper auto-correct :). Updated!

seroche:
I also took the liberty of upgrading the SDK version to latest in your `global.json`. Is this okay for you?

src/Serilog.AspNetCore/RequestLoggingOptions.cs
sungam3r: It means that it should have the same namespace to avoid writing additional `using` statement.

nblumhardt: Thanks @sungam3r; I meant the opposite, in this case - there's no need for a `using` statement even if the type is in a different namespace, since the members are accessed via the `options` action argument (sorry if I was unclear, it looks like the correct update has been made).

src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
```diff
@@ -12,24 +12,24 @@
...
 using Serilog.Parsing;
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
 
 namespace Serilog.AspNetCore
 {
-    class RequestLoggingMiddleware
+    internal class RequestLoggingMiddleware
```
nblumhardt: Nit: the Serilog projects don't use explicit `private` or `internal`

src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
```diff
@@ -68,13 +69,13 @@ public async Task Invoke(HttpContext httpContext)
             }
         }
 
-        bool LogCompletion(HttpContext httpContext, DiagnosticContextCollector collector, int statusCode, double elapsedMs, Exception ex)
+        private bool LogCompletion(HttpContext httpContext, DiagnosticContextCollector collector, int statusCode, double elapsedMs, Exception ex)
```
nblumhardt: Nit - explicit private

src/Serilog.AspNetCore/AspNetCore/RequestLoggingOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+        public string MessageTemplate { get; set; }
+
+        /// <summary>
+        /// Gets or sets the function returning the <see cref="LogEventLevel"/> based on the current <see cref="HttpContext"/>.
+        /// The default behavior returns LogEventLevel.Error when HttpStatusCode is greater than 499
+        /// </summary>
+        /// <value>
+        /// The function returning the <see cref="LogEventLevel"/>.
+        /// </value>
+        public Func<HttpContext, LogEventLevel> GetLogLevel { get; set; }
```
nblumhardt: `GetLevel` was the suggestion, here

src/Serilog.AspNetCore/SerilogApplicationBuilderExtensions.cs
```diff
@@ -26,6 +30,9 @@ public static class SerilogApplicationBuilderExtensions
         const string DefaultRequestCompletionMessageTemplate =
             "HTTP {RequestMethod} {RequestPath} responded {StatusCode} in {Elapsed:0.0000} ms";
 
+        static Func<HttpContext, LogEventLevel> DefaultGetLogLevel = 
+            ctx => ctx.Response.StatusCode > 499 ? LogEventLevel.Error : LogEventLevel.Information;
```
nblumhardt: `DefaultGetLevel` perhaps? Was the conversion to a method not pushed with the last changes?

src/Serilog.AspNetCore/RequestLoggingOptions.cs
sungam3r: ok

seroche:
Hopefully this should be good now :)

src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
```diff
@@ -71,10 +72,10 @@ public async Task Invoke(HttpContext httpContext)
         bool LogCompletion(HttpContext httpContext, DiagnosticContextCollector collector, int statusCode, double elapsedMs, Exception ex)
         {
             var logger = Log.ForContext<RequestLoggingMiddleware>();
-            var level = statusCode > 499 ? LogEventLevel.Error : LogEventLevel.Information;
+            var level = _getLevel(httpContext);
```
nblumhardt: I just realized that we've introduced a regression in the events raised via:

```
                when (LogCompletion(httpContext, collector, 500, GetElapsedMilliseconds(start, Stopwatch.GetTimestamp()), ex))
```

In this case, the 500 is being passed explicitly because it won't yet have been set on `HttpContext.Response.StatusCode`. There's a bit of awkwardness around this and we do sometimes lie about the status code because of this.

We could resolve this by making the signature of `_getLevel` accept both the context and the exception, and treat the level as 500 in `DefaultGetLevel` if the exception is present.

seroche: I had a similar problem in the past. An exception is raised but StatusCode is not yet set to 500. I changed the `Func` to accept the `HttpContext` and the `Exception`.  I'm tempted to create a custom context class exposing fewer properties. This context could be built before calling `_getLevel`. This class would contain a `HttpStatusCode` (500 if exception), the `Exception` (if any) and the Request full URL. It gives less flexibility than passing the HttpContext but at least we can control the values we pass to the Func. What do you think about that?

nblumhardt: I think `HttpContext` is a stable enough type that we can probably expose it as-is; otherwise we'll have to do a lot of guesswork to support the right scenarios.

Thanks for making those last few changes! I'll loop back to this and get it merged shortly.

seroche:
Hello, any plan to merge this soon?

nblumhardt:
Done! Thanks. I think if I get a chance I'll investigate adding `double elapsedMS` to the parameter list of `GetLevel`, since I expect some people will want to emit warnings if requests exceed a certain threshold.

PR #136, opened 2019-09-25
Update docs on disabling built-in logging

DavidCoghlan:
The current documentation recommends overriding the logging level for the entire Microsoft namespace, but this also suppresses logging for other Microsoft products, such as Entity Framework Core. Suggest updating this to recommend overriding Microsoft.AspNetCore instead.


nblumhardt:
Makes sense, thanks 👍

PR #137, opened 2019-10-02
Include 'double elapsedMs' in the function signature of the GetLevel delegate

nblumhardt:
Follows on from #132, includes request timing information in the level computation.


nblumhardt:
Sending this one along to include in the 3.1.0 prerelease

Issue #148, opened 2019-11-07
EnrichDiagnosticContext not adding properties to output


purplepangolin:
First, let me say thanks for providing this excellent set of projects. There are a lot of great features and impressive documentation of the capabilites provided. The following issue may be due to me misunderstanding something, but I _think_ it is a bug.

I  am using `Serilog.AspNetCore 3.2.0-dev-00133` as I wanted to take advantage of `EnrichDiagnosticContext` which has recently been added, in order to easily log request headers when using the `RequestLoggingMiddleware`.

```
    public void Configure(IApplicationBuilder app, IHostingEnvironment env)
    {
      app.UseSerilogRequestLogging(options =>
      {
        options.MessageTemplate = MessageTemplate;
        options.EnrichDiagnosticContext = (diagnosticContext, httpContext) =>
        {
          diagnosticContext.Set("Headers", httpContext.Request.Headers);
        };
      });
      
      app.UseMvc();
    }
```

However, the headers do not appear in the logged event. Looking at the code for `RequestLoggingMiddleware.LogCompletion`  Lines 81-85 are:

```
            if (!collector.TryComplete(out var collectedProperties))
                collectedProperties = NoProperties;

            // Enrich diagnostic context
            _enrichDiagnosticContext?.Invoke(_diagnosticContext, httpContext);
```

shouldn't the call to enrich appear before `collector.TryComplete` as per [this comment](https://github.com/serilog/serilog-aspnetcore/pull/146#issuecomment-548112138)?

nblumhardt:
Fixed in #149, thanks 👍

PR #149, opened 2019-11-07
FIX: Enrich context before collecting properties

purplepangolin:
LogCompletion event was not including properties added via
`EnrichDiagnosticContext` delegate on options because
the properties had been collected before invoking the enricher.

Swapping the order of the statements fixes this.

This is a fix for https://github.com/serilog/serilog-aspnetcore/issues/148

I'm afraid I haven't had time to add a test for this. I have built the package locally and tested it with the project I am currently working on.

### DIFF
--- a/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
+++ b/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
@@ -78,11 +78,11 @@ namespace Serilog.AspNetCore
 
             if (!logger.IsEnabled(level)) return false;
 
-            if (!collector.TryComplete(out var collectedProperties))
-                collectedProperties = NoProperties;
-
             // Enrich diagnostic context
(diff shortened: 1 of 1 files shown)


nblumhardt:
Oops! Nice catch, thanks 👍

D9001:
Hi, I was following some blog posts and set up my app to use `EnrichDiagnosticContext` from within `UseSerilogRequestLogging` in a .NET Core 2.0 library. This made my app logging so much cleaner. 
However, I just realized that I only have access to these when I use I use one of the latest dev versions of the nuget package. Is there going to be a stable release available soon which makes use of these? Just wanted to make sure before I roll back my changes in time for our production release.

nblumhardt:
@D9001 I'll set up a WIP release PR now so that you can track it, thanks; sounds like you're happy with the shape of `EnrichDiagnostiContext` as it's implemented in the previews?

D9001:
@nblumhardt Absolutely. Thanks!

PR #165, opened 2020-01-18
Update to lts 31

tebeco:
This PR intention is to update it to `netcoreapp3.1` as it is LTS.
As of today, when `serilog-aspnetcore` is used in project targeting `netcoreapp3.1`, it "leaks" few transitive dependencies `Microsoft.AspNetCore.xxx` and `Microsoft.Extensions.xxx` version `2.x.y`.

* Update dependencies from `2.0.x` to `2.1.*` (Wildcard to get the latest patch available at build time)
* As `AspNetCore 3.1` is NOT `netstandard2.0/2.1` but only `netcoreapp3.1`, i cross targetted to both
  * `netcoreapp3.1` for `AspNetCore3.1` (LTS)
  * `netstandard2.0` for `AspNetCore2.1` (Also LTS)
* As `AspNetCore 3.1` is LTS, i changed the Sample `TargetFramework` from `netcoreapp3.0` to `netcoreapp3.1`
* I've set `MsBuild` Condition on the `TargetFramework` because 
  * `AspNetCore2.x` is build with direct dependencies on `Msft.Extensions 2.x`
  * `AspNetCore2.x` is build with direct dependencies on `Msft.Extensions 3.x`
  This means the dependencies version need to follow this
* I change `AppVeyor` image to `Vs 2019` (because netcoreapp is supposed to requires vs2019 but not sure it's needed here are running `dotnet-install.ps1 -Channel 3.1` would just install the latest `3.1 SDK`
* I removed all `powershell` warning from the linter (general idea is to never use alias, as it could leads to missunderstand or error, eg : `rm -rf` the `-f` is for `-Filter` and not `--force`)
* added `serilog-extension-nuget.png` to the repo and `PackageIcon` as the current `PackageIconUrl` is is deprecated for recent SDK (see https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packageiconurl)


tebeco:
I did not checked if some of the `Serilog` dependencies could be updated here to be fair :
```
<ItemGroup>
    <PackageReference Include="Serilog" Version="2.8.0" />
    <PackageReference Include="Serilog.Extensions.Hosting" Version="3.0.0" />
    <PackageReference Include="Serilog.Sinks.Console" Version="3.1.1" />
    <PackageReference Include="Serilog.Sinks.File" Version="4.0.0" />
    <PackageReference Include="Serilog.Sinks.Debug" Version="1.0.1" />
    <PackageReference Include="Serilog.Settings.Configuration" Version="3.1.0" />
    <PackageReference Include="Serilog.Formatting.Compact" Version="1.0.0" />
</ItemGroup>
```
but i can take a look

tebeco:
as you can see here :
image
* `Serilog.Extension.Hosting` need to same update because of `Microsoft.Extensions.Hosting.Abstractions (2.1.0)`
* `Serilog.Settings.Configuration` for `Microsoft.Extensions.Options.ConfigurationExtensions (2.0.0)` and `Microsoft.Extensions.DependencyModel (2.0.4)`

Build.ps1
```diff
@@ -1,24 +1,24 @@
...
-	echo "build: Cleaning .\artifacts"
+	Write-Output "build: Cleaning .\artifacts"
 	Remove-Item .\artifacts -Force -Recurse
 }
 
 & dotnet restore --no-cache
 
-$branch = @{ $true = $env:APPVEYOR_REPO_BRANCH; $false = $(git symbolic-ref --short -q HEAD) }[$env:APPVEYOR_REPO_BRANCH -ne $NULL];
-$revision = @{ $true = "{0:00000}" -f [convert]::ToInt32("0" + $env:APPVEYOR_BUILD_NUMBER, 10); $false = "local" }[$env:APPVEYOR_BUILD_NUMBER -ne $NULL];
+$branch = @{ $true = $env:APPVEYOR_REPO_BRANCH; $false = $(git symbolic-ref --short -q HEAD) }[$null -ne $env:APPVEYOR_REPO_BRANCH];
```
sungam3r: What is the reason for changing the order of the arguments?

tebeco: it's a warning from the linter too, shipped within the `Powershell` extensions (msft powershell team)

i don't have the underlying reason for this one.
my personal guess would be that it applies the same principle as some unit test lib : expected/constants on the left operand.

but it's just a guest

tebeco:
FYI : currently updating the 2 other serilog repository that leak transitive from 2.x and i will really need a decision from `serilog maintainer`
more detail in these PR

Build.ps1
sungam3r: This reminds me https://en.wikipedia.org/wiki/Yoda_conditions

tebeco: i can totally revert it ;)
i just wanted to get rid of warning to be fair ^^

sungam3r: It doesn’t matter to me.

andrewlock:
I previously suggested multi-targeting to 3.x with @nblumhardt in https://github.com/serilog/serilog-extensions-logging/pull/155 but we concluded there's actually no reason to do so, and it makes tooling etc more complicated.

I discussed the issues recently in a blog post: https://andrewlock.net/converting-a-netstandard-2-library-to-netcore-3/#libraries-that-depend-on-microsoft-extensions-nuget-packages

The conclusion was that there's no need to upgrade this library, serilog-extensions-hosting, or serilog-extensions-logging! The other commits that remove deprecated properties etc would definitely be handy though 🙂

tebeco:
because `serilog-extensions-logging` only consume `Microsoft.Extensions.Logging`
https://github.com/serilog/serilog-extensions-logging/blob/dev/src/Serilog.Extensions.Logging/Serilog.Extensions.Logging.csproj#L32

Thought this repo consumes a package the was previously shipped as a Nuget, and now from the `FrameworkReference` / SDK

tebeco:
As you said, yes i removed cross targeting to `netcoreapp3.1` for the 2 other repo few minutes ago ;)

Also, as explained earlier, there's only 2 LTS version that has not reached End Of Life yet : 2.1 and 3.1
So anybody staying on 2.1 kinda "pinned down" his runtime already.
My point is that, pinning the runtime also means pinning dependencies compliant with that runtime.

This is done already in `Serilog.Settings.Configuration`, for example :
https://github.com/serilog/serilog-settings-configuration/blob/dev/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj#L34

The current

tebeco:
EDIT : previous answers was weird ;)

As you said, `serilog-extensions-logging` only consume `Microsoft.Extensions.Logging`
https://github.com/serilog/serilog-extensions-logging/blob/dev/src/Serilog.Extensions.Logging/Serilog.Extensions.Logging.csproj#L32
So you are right `serilog-extensions-logging`, can stay on `netstandard2.0`

This repo consumes a package the was previously shipped as a Nuget, and now shipped from the `FrameworkReference` (which is only `netcoreapp3.1`)
https://github.com/serilog/serilog-aspnetcore/blob/dev/src/Serilog.AspNetCore/Serilog.AspNetCore.csproj#L31


I made 2 other PR to 2 others repo that only consumed `Microsoft.Extension.XXX`, 
I first used `netcoreapp3.1` to discriminates for consumer that uses `AspNetCore 2.x` or `AspNetCore 3.x`

Few minutes before your answer here, i rolled that back to `netstandard2.0` with this idea :
There's only 2 LTS version that has not reached End Of Life yet : 2.1 and 3.1
So anybody staying on 2.1 kinda is "pinning his runtime" to a specific (old/older) version already.
My point is that, pinning the runtime also means pinning dependencies compliant with that runtime.

This is done already in Serilog.Settings.Configuration, for example :
https://github.com/serilog/serilog-settings-configuration/blob/dev/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj#L34

nblumhardt:
Hi @tebeco - I think we should go ahead with this change; thanks for sending the PR, and sorry about the slow turnaround at this end.

Would it be possible to remove the other build script and CSPROJ changes (latest version references etc.) from the PR, and just commit the `netcoreapp3.1` targeting change? The other parts are somewhat standardized across all of the Serilog repos and any changes there will need more consideration/discussion that's separate to the issue covered by this PR.

Thanks!

tebeco:
So this means that when the nuget is released, serilog transitive will potentially points to deprecated version ?

tebeco:
As you said, you standardize few things, like the icon URL. As nuget team is actively depracting it, that's why i changed it here as it's probably going to be needed.

image

No idea why the change, but i guess that relying on an external URL for an icon can lead to issues like DNS issues / timeout / unpaid domain / overflow / crafted response / ...

tebeco:
@nblumhardt 
Can you check the new history now ? Does it look like what you had in mind ?

nblumhardt:
Thanks for the follow-up @tebeco :+1: 

We should make the `<PackageIcon>` change - sorry I wasn't clear (we've updated this in many other Serilog repos already). It was changes like `$(LatestVersion)` that I'd prefer to hold off.

It looks like you may not have pushed your updated changes to your branch? Everything still looks the same. Cheers!

tebeco:
@nblumhardt 
Sorry I pushed the wrong branch -_-
classic PEBKAC

src/Serilog.AspNetCore/Serilog.AspNetCore.csproj
```diff
@@ -21,16 +22,30 @@
...
     <PackageReference Include="Serilog.Settings.Configuration" Version="3.1.0" />
-    <PackageReference Include="Serilog.Formatting.Compact" Version="1.0.0" />
-    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.0.0" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.0.0" />
+    <PackageReference Include="Serilog.Formatting.Compact" Version="1.1.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.1.*" />
```
nblumhardt: `*` wildcards get baked into the package as specific versions at the time of package build (not package install), and this has caused a bit of confusion historically. Because they make builds non-deterministic we have chosen to avoid them in other Serilog projects. Can we please use specific versions, here?

tebeco: i'm fine with both, wildcard, explicit latest or explicit 0 ?
both can be confusing to it's up to the maintainer to make a choice

as a very personal and opinionated, i find it useful to have this behavior so you just rebuild libs when needing for an explicit roll forward and having a release (even though being open on en right "should" be enough)
it make it easier to understand against which version it is as built => as you said the constraint on the package being installed are the only source of truth.

i encountered the opposite side effect ^^
building against 2.1.0 instead of 2.1.* confused user saying that it was built against potential bug fixed in later release ;)
i would then try to suggest "explicit latest" here

what's your call ?

will push the fix tomorrow regardless of 2.1.0 or 2.1.(can't remember the actual digit)

nblumhardt: Thanks for the follow-up :-)  - yes, the trade-off is an annoying one. For now I think being consistent with the rest of the Serilog projects is worthwhile.

I think your "explicit latest" suggestion is a good option, let's go with that 👍

tebeco: perfect I was opening nuget.org right now to anticipate that ^^

will be pushed in like ~5min

tebeco: done

tebeco:
@nblumhardt 
any thing else I could have missed ?

tebeco:
regarding all the `ps1` stuff, it almost feels like you want to create a `Serilog.BuildTool` package just like `dotnet/arcade` so that all your technical scripts / images for the `Icon`  / `Copyright` etc ... are all in an `eng` folder (engineering I guess)

so everytime you update the build process,
* you update that lib (or `dotnet tool` ?)
* pull it down per repo
* ran it so that it update `eng` folder
* you're never supposed to manually edit `eng` folder (like a lock file)
* `git add eng && git commit -m "update build tooling"`

that way you could may a bot attempts to automatically creates PR + run the script of all organization repositories here 

=> One fix to fix them all

tebeco:
(rebased / force-with-lease on `dev` branch)

nblumhardt:
:+1: thanks! I'll bump the package version to 3.3.0 because of the dependency version updates. We'd normally consider these a breaking change, but since the package will still run on the earlier versions of those dependencies, I think we can squeeze this into a minor version bump.

tebeco:
Thx a lot for your time ;)

Do you think the others repo that were not bump could cross target `netstandard2.1`, i understand that they do no explicitly uses `ns2.1` API, but that would makes `netcoreapp3.1` resolves it, so that Transtive like `Msft.Ext.Logging` could be resolving `3.1.2` for example.

So far nuget resolves the oldest possible match => it does not respect anymore `HigestDependency` Settings so sometimes it resolves `Msft.Ext.Logging 2.1` at work for stacks using `netcoreapp3.1`

That would be a way to helps nuget resolves proper dependencies based on the TFM wihtout breaking backward compat.
Should i re-open an issue about that on these repo ?
(pinging @andrewlock as he was also involved with you in that process / discussion)

CleverNeologism:
Should the SDK in global.json also be updated to latest 3.1?  Stock VS2019 install will not open this solution, and the current.

tebeco:
for this PR ?
image

tebeco:
https://github.com/serilog/serilog-aspnetcore/pull/172

sungam3r:
@CleverNeologism @tebeco Hi, can you tell me what is the point in the global.json file? When I do not have the exact sdk mentioned in global.json, then solution does not open. It’s easier for me just to delete this file so that everything works.

tebeco:
https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x

PR #166, opened 2020-01-23
Add LoggerConfiguration in IServiceCollection

sungam3r:
@nblumhardt I found that I can solve the problem I described in #130 much easier. I just need to add `LoggerConfiguration` to the container. I see that this is already happening with `IDiagnosticContext` `DiagnosticContext` and `ILogger` so I think there will be nothing wrong with adding `LoggerConfiguration`. Also note that I use `TryAddSingleton` to avoid potential problems with already added configuration.

### DIFF
--- a/src/Serilog.AspNetCore/SerilogWebHostBuilderExtensions.cs
+++ b/src/Serilog.AspNetCore/SerilogWebHostBuilderExtensions.cs
@@ -17,6 +17,7 @@ using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Logging;
 using Serilog.Extensions.Logging;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Serilog.Extensions.Hosting;
 
 namespace Serilog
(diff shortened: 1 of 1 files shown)


nblumhardt:
Hi @sungam3r - thanks for the PR!

I'm not confident this is the right direction, since it's very order-dependent (`LoggerConfiguration` can't be used for anything once `CreateLogger()` has been called) and unclear where it would fit into typical use, especially when #130 only appears to be an issue for pre-generic-host code, and had a good workaround.

Am I missing a scenario, here? Thanks again.

sungam3r:
> `LoggerConfiguration` can't be used for anything once CreateLogger() has been called

I use my own components use which use `LoggerConfiguration` from DI. Yes, I create my own loggers that rely on the `LoggerConfiguration`. But they are created from DI, implicitly by the calling code, which does not directly refer to the Serilog interfaces. I have an abstract logger that can be resolved as a Serilog or NLog logger or other. In case of Serialog, the constructor parameter is needed - `LoggerConfiguration` since it contains all the information necessary for the logger.

sungam3r:
Abstracting from a specific application library of logging allowed us to smoothly migrate from NLog to Serilog some time ago without changing the numerous libraries that used logging primitives. All that was required was to set a different implementation in DI at the projects Composition Roots. We have a microservice architecture with hundreds of independent services / deployments. Therefore, we design the system as independent as possible from external entities so that it is not necessary to recompile the lower and middle layers of applications - only change Composition Root configuration.

sungam3r:
Relates to serilog/serilog#1358

sungam3r:
Nevermind, I can use `ILogger` directly in my ctors instead of configuration.

PR #167, opened 2020-01-23
Update Instructions for .NET Core 3 templates

### LINKED ISSUE #147: Issue with this: Remove ILoggerFactory parameters and any Add*() calls on the logger factory in Startup.cs
Hi, I'm wanting to use the serilog-aspnetcore package and am following instructions in the readme file but am stumbling on this line

> Remove ILoggerFactory parameters and any Add*() calls on the logger factory in Startup.cs

I have this line in my startup.cs for registering a custom modelbinderprovider 

```
services.AddMvc().AddMvcOptions(options => { 
  options.ModelBinderProviders.Insert(0, new GenericViewModelBinderProvider(loggerFactory));
})"
```

I need to pass an ILoggerFactory because GenericViewModelBinderProvider creates a new type of modelbinder which inherits from ComplexTypeModelBinder, which needs an ILoggerFactory. 

My question is why do we need to remove ILoggerFactory parameters from the Startup.cs file. It would seem to me that if  and what is the best method for getting the serilog logger factory within the configureservices method?

I hope this makes sense. Thank you for your time.

phlashdev:
Fixes #147 

The instructions for using Serilog contained statements that were no longer valid since .NET Core 3.0 uses the Generic Host and new project templates.

The code sample in the readme and the corresponding statements were updated to resemble these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ public class Program
         try
         {
             Log.Information("Starting web host");
-            CreateWebHostBuilder(args).Build().Run();
+            CreateHostBuilder(args).Build().Run();
             return 0;
         }
         catch (Exception ex)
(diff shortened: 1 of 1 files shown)


phlashdev:
I think that probably it would be a good idea to inlcude the instructions for users that are still on pre .NET Core 3 versions. But as it seems to me the readme file resembles the instructions for the latest version, and as the templates vary also for example from .NET Core 2.0 to 2.1, this could make the readme somewhat unclear? 

I'm a little unsure about we should do and how we do it, so I really would like to know what you're thinking.

nblumhardt:
I think it's worthwhile moving the examples forwards, since 3.x is now well-and-truly established. Thanks!

Issue #168, opened 2020-02-05
Yet another middleware request/response logging question


vladimirvozar:
Hi,
I'm trying to implement request/response logging middleware that would be logging request and response body payload as serilog properties (logging to .json file).

For GET requests, body of the response should be logged (added to serilog context as a property), and for POST requests, both body of request and body of response should be logged. 
I have created middleware and managed to properly retrieve data from request and response stream, and getting it as a string, but only the "RequestBody" is being logged properly.

When debugging, I can see that reading request/response body works fine.

Following is the code excerpt from Program->Main method:
```
Log.Logger = new LoggerConfiguration()
    .ReadFrom.Configuration(configuration)
    .Enrich.FromLogContext()
    .CreateLogger();
```
and code in the middleware:
```
public async Task Invoke(HttpContext context)
{
    // Read and log request body data
    string requestBodyPayload = await ReadRequestBody(context.Request);

    LogContext.PushProperty("RequestBody", requestBodyPayload);

    // Read and log response body data
    var originalBodyStream = context.Response.Body;
    using (var responseBody = new MemoryStream())
    {
        context.Response.Body = responseBody;
        await _next(context);
        string responseBodyPayload = await ReadResponseBody(context.Response);

        if (!context.Request.Path.ToString().Contains("swagger.json") && !context.Request.Path.ToString().Contains("index.html"))
        {
            LogContext.PushProperty("ResponseBody", responseBodyPayload);
        }

        await responseBody.CopyToAsync(originalBodyStream);
    }
}

private async Task<string> ReadRequestBody(HttpRequest request)
{
    HttpRequestRewindExtensions.EnableBuffering(request);

    var body = request.Body;
    var buffer = new byte[Convert.ToInt32(request.ContentLength)];
    await request.Body.ReadAsync(buffer, 0, buffer.Length);
    string requestBody = Encoding.UTF8.GetString(buffer);
    body.Seek(0, SeekOrigin.Begin);
    request.Body = body;

    return $"{requestBody}";
}

private async Task<string> ReadResponseBody(HttpResponse response)
{
    response.Body.Seek(0, SeekOrigin.Begin);
    string responseBody = await new StreamReader(response.Body).ReadToEndAsync();
    response.Body.Seek(0, SeekOrigin.Begin);

    return $"{responseBody}";
}
```
As I mentioned, "RequestBody" is properly logged to file (POST requests) , but nothing for "ResponseBody" (not even added as a property).
Appreciate any help

[...]

PR #173, opened 2020-02-19
Add async version of diagnostic context enricher callback.

CleverNeologism:
Reasoning:

Accessing the request or response streams to potentially include them in these logs in ASP.NET Core 3.1 is problematic using the existing call, since [3.1 doesn't want you to access these synchronously](https://github.com/dotnet/aspnetcore/issues/7644).  Forcing a global option to be set or wrapping the async calls in the existing sync callback for this seems inelegant.  Provide an async version for the callback in a non-breaking fashion as an option.  Perhaps mark sync version as obsolete and move to async in the future, as middleware framework is async?

### DIFF
--- a/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
+++ b/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
@@ -30,6 +30,7 @@ namespace Serilog.AspNetCore
         readonly DiagnosticContext _diagnosticContext;
         readonly MessageTemplate _messageTemplate;
         readonly Action<IDiagnosticContext, HttpContext> _enrichDiagnosticContext;
+        readonly Func<IDiagnosticContext, HttpContext, Task> _enrichDiagnosticContextAsync;
         readonly Func<HttpContext, double, Exception, LogEventLevel> _getLevel;
         static readonly LogEventProperty[] NoProperties = new LogEventProperty[0];
 
@@ -41,6 +42,7 @@ namespace Serilog.AspNetCore
 
             _getLevel = options.GetLevel;
             _enrichDiagnosticContext = options.EnrichDiagnosticContext;
+            _enrichDiagnosticContextAsync = options.EnrichDiagnosticContextAsync;
             _messageTemplate = new MessageTemplateParser().Parse(options.MessageTemplate);
         }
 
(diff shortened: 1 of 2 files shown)


src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
```diff
@@ -57,29 +59,31 @@ public async Task Invoke(HttpContext httpContext)
...
                 var statusCode = httpContext.Response.StatusCode;
-                LogCompletion(httpContext, collector, statusCode, elapsedMs, null);
+                await LogCompletion(httpContext, collector, statusCode, elapsedMs, null);
             }
             catch (Exception ex)
-                // Never caught, because `LogCompletion()` returns false. This ensures e.g. the developer exception page is still
-                // shown, although it does also mean we see a duplicate "unhandled exception" event from ASP.NET Core.
-                when (LogCompletion(httpContext, collector, 500, GetElapsedMilliseconds(start, Stopwatch.GetTimestamp()), ex))
             {
+                await LogCompletion(httpContext, collector, 500, GetElapsedMilliseconds(start, Stopwatch.GetTimestamp()), ex);
```
nblumhardt: This is a regression - the `when` clause executes before unwinding, which we use in order to capture `LogContext` state before it's popped off as `using` blocks are exited.

nblumhardt:
Thanks for the PR!

I think it may be preferable to use a workaround like the one in https://github.com/serilog/serilog-aspnetcore/issues/168#issuecomment-582896813 - first, because we want to continue using exception filters to capture local context when an exception is thrown, but also, because even with the callback, it's very hard to reliably include request/response stream contents. A separate middleware project for this purpose might be worth exploring, but currently I think it's out of scope for this library.

Cheers,
Nick

CleverNeologism:
@nblumhardt 

> A separate middleware project for this purpose might be worth exploring, but currently I think it's out of scope for this library.

I would think optionally including request/response data (as much as possible, anyway) would be a good feature to include since you are logging request information, especially since, as you say, it can be tricky (especially the response).  I might submit another PR with it, as I've written this code several times to capture requests that generate errors.

Also, learned some interesting minutiae about when clauses from your comment.  That pretty much kills making the callback async, since you can't await in a when clause.  Thanks for the detailed response.

PR #178, opened 2020-02-27
Remove unnecessary dependencies

inosik:
This removes packages that are used in the sample projects and some unused packages from the actual _Serilog.AspNetCore_ package.


sungam3r:
Well, see #127 :)

inosik:
> Wow! I could not even imagine that it was intended.

+1 😄 

Well, @nblumhardt feel free to close this then.

sungam3r:
I still have a negative attitude towards the decision to include these dependencies in the package.

nblumhardt:
Thanks all. I think the original logic is still sound:

> something resembling a meta-package is needed if we're going to meet the ease-of-setup expectations set by Microsoft.Extensions.Logging/ASP.NET Core, which includes console and debug logging by default.

but we might meet the same goals by making _Serilog.AspNetCore_ a pure metapackage in the future. Worth considering, anyway!

PR #183, opened 2020-03-25
Allow an ILogger to be specified

### LINKED ISSUE #182: Middleware relies on static Log.Logger being set first.
I am trying to use only instances of ILogger. As such I wondered what would happen if I never set `Log.Logger.`

I found was now getting no request logging. Looking at the request logging middleware it looks like Log.Logger needs to be set first as it uses Log.ForContext: https://github.com/serilog/serilog-aspnetcore/blob/dev/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs#L76 and if Log.Logger isn't set this yields SilentLogger's. (To be honest I would think it would be better to throw an exception in such a case as logging not being initialised is pretty important)

I'd propose an enhancement to the middleware so that an instance of a logger can be provided and that used to get log context instead --> falling back to the static Log.Logger if no instance provided might be a good way to keep this a backwards compatible change?

nblumhardt:
Resolves #182

### DIFF
--- a/src/Serilog.AspNetCore/AspNetCore/RequestLoggingOptions.cs
+++ b/src/Serilog.AspNetCore/AspNetCore/RequestLoggingOptions.cs
@@ -16,6 +16,8 @@ using Microsoft.AspNetCore.Http;
 using Serilog.Events;
 using System;
 
+// ReSharper disable UnusedAutoPropertyAccessor.Global
+
 namespace Serilog.AspNetCore
 {
     /// <summary>
@@ -49,6 +51,13 @@ namespace Serilog.AspNetCore
         /// A callback that can be used to set additional properties on the request completion event.
         /// </summary>
         public Action<IDiagnosticContext, HttpContext> EnrichDiagnosticContext { get; set; }
+
+
+        /// <summary>
+        /// The logger through which request completion events will be logged. The default is to use the
+        /// static <see cref="Log"/> class.
+        /// </summary>
+        public ILogger Logger { get; set; }
 
         internal RequestLoggingOptions() { }
     }
--- a/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
+++ b/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
@@ -24,6 +24,7 @@ using System.Threading.Tasks;
 
 namespace Serilog.AspNetCore
 {
+    // ReSharper disable once ClassNeverInstantiated.Global
     class RequestLoggingMiddleware
     {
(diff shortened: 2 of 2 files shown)


src/Serilog.AspNetCore/AspNetCore/RequestLoggingOptions.cs
```diff
@@ -50,6 +52,13 @@ public class RequestLoggingOptions
         /// </summary>
         public Action<IDiagnosticContext, HttpContext> EnrichDiagnosticContext { get; set; }
 
+
+        /// <summary>
+        /// The logger through which request completion events will be logged. The default is to use the
+        /// static <see cref="Log"/> class.
+        /// </summary>
+        public ILogger Logger { get; set; }
```
sungam3r: https://github.com/serilog/serilog-aspnetcore/blob/dev/src/Serilog.AspNetCore/SerilogWebHostBuilderExtensions.cs#L146

I think it is worth to use DI provided features. Why to manually specify logger if it is already exists ? Just add new ctor param:
```c#
 public RequestLoggingMiddleware(RequestDelegate next, DiagnosticContext diagnosticContext, RequestLoggingOptions options, ILogger logger)
```

dazinator: This won't work if serilog ILogger is not registered for DI and somneone is only using Log.Logger everywhere - which i may be wrong, but is a valid style for using serilog.

sungam3r: Well, then we can use second RequestLoggingMiddleware ctor with ILogger property inside options to satisfy both categories.

nblumhardt: Thanks for the feedback @sungam3r - will give it some thought :+1:

pnmcosta:
Any prediction when this might be merged? Thanks.

nblumhardt:
@pnmcosta still needs some thought, sorry. I like the idea of accepting the logger as a constructor parameter, but I'm not sure whether this covers all scenarios 🤔

pnmcosta:
@nblumhardt my scenario is that we build a library that extends Serilog with our own conventions, and we expose our own singleton for the root logger, not relying on the Serilog's.

We also only leverage DI with MS's ILogger's, i'm not even sure if internally Serilog.Ilogger's are registered in the container, but in our case we don't need the Serilog.ILogger in the container.

From a consuming point of view, I'd definitely be more inclined to set the logger on the options, unless `UseSerilog` does it for us, as I am setting our singleton there.

HTH

PR #191, opened 2020-04-30
Fallback RequestPath for when IHttpRequestFeature.RawTarteg is empty

rsantosdev:
When running integration tests using the WebApplicationFactory class, I noticed that the code `httpContext.Features.Get<IHttpRequestFeature>()?.RawTarget` returns an empty string instead of null, which causes the fallback never executing and displaying no path information.


nblumhardt:
Thanks! Since it's a bit of an odd non-mainline scenario (I can't help feeling that the test harness is broken, here), perhaps we go for the even more tactical `string.IsNullOrEmpty()` and add a comment explaining the involvement of `WebApplicationFactory`?

rsantosdev:
@nblumhardt thanks for the feedback. I added request adjustments

nblumhardt:
Thanks! :+1:

PR #193, opened 2020-05-11
Update to show new UseSerilog() overload including service provider

nblumhardt:
Shows the new `UseSerilog()` inline overload from https://github.com/serilog/serilog-extensions-hosting/pull/20

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "sdk": {
     "allowPrerelease": false,
-    "version": "3.1.100",
+    "version": "3.1.201",
     "rollForward": "latestPatch"
   }
 }
--- a/global.json
+++ b/global.json
@@ -2,6 +2,6 @@
   "sdk": {
     "allowPrerelease": false,
     "version": "3.1.201",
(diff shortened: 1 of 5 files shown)


global.json
```diff
@@ -1,7 +1,7 @@
 {
   "sdk": {
     "allowPrerelease": false,
-    "version": "3.1.100",
+    "version": "3.1.201",
     "rollForward": "latestPatch"
   }
```
skomis-mm, line 6: probably better once set to a [latestFeature](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x#matching-rules)?

sungam3r: Or just remove `global.json`

nblumhardt: Definitely can be overkill for many projects 👍  - for the core _Serilog_ package and a couple of other sensitive ones like this, I think the improved determinism is worth the inconvenience, but outside of this handful of packages I've tried to avoid it :-)

nblumhardt:
Thanks

Issue #207, opened 2020-08-03
Opt-in to TraceId and SpanId enrichment for ASP.NET Core 5.0 logs


cloudnin9:
Hello, Serilog community. Firstly, thank you for the great library!

**Description**
I am trying out Asp.Net on Net 5 Preview 7 so I have created 2 web API projects one targeting [netcoreapp3.1] and another targeting [net5]

The problem is when I am looking at logs, I no longer see TraceId and SpanId for the API targeting [net5] with the DotNet 5 preview 7 SDK.

I also tested using the vanilla Logger and there was no issue there. Do I have to configure something, did I just miss something or is the DotNet 5 Preview SDK not fully supported yet?

Thanks for any info.

**Reproduction**
Below is my bootstrapping code, it is identical for both APIs,
```
using System.Diagnostics;
using Microsoft.AspNetCore.Hosting;
using Microsoft.Extensions.Hosting;
using Serilog;
using Serilog.Formatting.Elasticsearch;

namespace WebApplication1
{
    public class Program
    {
        public static void Main(string[] args)
        {
            Activity.DefaultIdFormat = ActivityIdFormat.W3C;
            Log.Logger = new LoggerConfiguration()
                .MinimumLevel.Debug()
                .Enrich.FromLogContext()
                .WriteTo.Console(new ExceptionAsObjectJsonFormatter(renderMessage: true))
                .CreateLogger();
            CreateHostBuilder(args).Build().Run();
        }

        public static IHostBuilder CreateHostBuilder(string[] args) =>
            Host.CreateDefaultBuilder(args)
            .UseSerilog()
                .ConfigureWebHostDefaults(webBuilder =>
                {
                    webBuilder.UseStartup<Startup>();
                });
    }
}
```
**Expected behavior**
Be able to see TraceId and SpanId in my logs, regardless of the AspDotnet runtime version

**Relevant package, tooling and runtime versions**
```
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
    <UserSecretsId>b7a811b1-eb5a-4655-b19c-5e97c4ae1370</UserSecretsId>
    <DockerDefaultTargetOS>Linux</DockerDefaultTargetOS>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.10.9" />
[...]

PR #215, opened 2020-10-22
Add some request headers to the log

SimonCropp:
The reason it does a hard coded subset is i want to avoid writing security info that may be in other headers (eg cookies).

### DIFF
--- a/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
+++ b/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
@@ -18,6 +18,7 @@ using Serilog.Events;
 using Serilog.Extensions.Hosting;
 using Serilog.Parsing;
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
```diff
@@ -104,7 +107,7 @@ bool LogCompletion(HttpContext httpContext, DiagnosticContextCollector collector
 
         static double GetElapsedMilliseconds(long start, long stop)
         {
-            return (stop - start) * 1000 / (double)Stopwatch.Frequency;
+            return (stop - start) * 1000 / (double) Stopwatch.Frequency;
```
sungam3r, line 110:
```suggestion
            return (stop - start) * 1000 / (double)Stopwatch.Frequency;
```

src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
```diff
@@ -119,8 +122,31 @@ static string GetPath(HttpContext httpContext)
...
+            "content-length",
+            "content-type",
+            "referer",
+            "user-agent"
+        };
+
+        static LogEventProperty GetRequestHeaders(HttpContext httpContext)
+        {
+            var values = httpContext.Request.Headers
+                .Where(x => allowedHeaders.Contains(x.Key.ToLowerInvariant()))
```
sungam3r, line 144: Maybe use case insensitive comparoson instead `ToLower` ?

nblumhardt:
Hi @SimonCropp!

This would need to be an opt-in; currently the middleware adds only what's required in order to render its message template, since the bulk of request logs can quickly climb once more details are included.

#207 is related; we might want to align them somewhat. `UseSerilogRequestLogging(opts => ...)` gives us a hook point for something like `opts.EnrichFromActivity= (an ActivityTrackingOptions flag set)` and `opts.EnrichFromRequest = (..something)`...?

PR #234, opened 2021-02-14
Update to include CreateBootstrapLogger()

nblumhardt:
**Breaking change**, the upstream _Serilog.Extensions.Hosting_ from this version only supports targets compatible with `netstandard2.1`, which this PR now adopts into this package as well. .NET Framework and earlier .NET Core consumers will need the 3.1.0 version of this package.

See https://nblumhardt.com/2020/10/bootstrap-logger/ for a discussion of the new hosting features.

See https://github.com/serilog/serilog-extensions-hosting/pull/35 for the corresponding _Serilog.Extensions.Hosting_ release.

This PR:

 * Bump major version
 * Target `netstandard2.1` only
 * Update README to show simplified examples with two-stage initialization
 * New `net5.0` MVC sample with two-stage initialization
 * Removed the separate early/inline intialization samples
 * Fix the embedded package icon


nblumhardt:
Ready to go :-) 👍

nblumhardt:
[README, rendered](https://github.com/serilog/serilog-aspnetcore/blob/2f449ddf76e730ef61b6be2d1a2187c2225a6367/README.md)

nblumhardt:
Sending this through to `dev`.

PR #247, opened 2021-05-27
Add using statement to readme example

acohenOT:
`LogEventLevel.Information` requires `using Serilog.Events;`


nblumhardt:
Thanks!

Issue #270, opened 2021-11-04
Provide exception in IDiagnosticContext


angularsen:
**Is your feature request related to a problem? Please describe.**
We use [Hellang.Middleware.ProblemDetails](https://www.nuget.org/packages/Hellang.Middleware.ProblemDetails/) to transform unhandled exceptions to ProblemDetails responses.

This means Serilog's `RequestLoggingMiddleware` does not observe any unhandled exceptions.

In order to log unhandled exceptions in our APIs, we currently use a separate error logging middleware.
This means we get two log events:
- One from a custom ErrorLogMiddleware.
- One from Serilog's request log middleware.

We would like to merge the exception data into Serilog's request log event by passing on the exception details via `IDiagnosticContext`.

**Describe the solution you'd like**
- Add `IDiagnosticContext.SetException(Exception e)`.
- Add `DiagnosticContextCollector.TryComplete(out IEnumerable<LogEventProperty>, out Exception)`
- Enrich log event with exception from `IDiagnosticContext`, unless an unhandled exception takes precedence.

**Pull requests**
https://github.com/serilog/serilog-extensions-hosting/pull/56
https://github.com/serilog/serilog-aspnetcore/pull/271

**Describe alternatives you've considered**
Separate error logging middleware, but then we get two log events instead of one.

**Additional context**
Current state:
![image](https://user-images.githubusercontent.com/787816/140333572-8f0a3555-1eb5-4d0b-8e07-6bb134510a45.png)

Desired state with exception merged into Serilog's request log event:
![image](https://user-images.githubusercontent.com/787816/140333867-2466ae42-9918-4e29-9ab2-07ab72fa9483.png)

nblumhardt:
Thanks for the PR and write-up 👍 

Sounds like a reasonable solution; will need to give this some thought but will loop back to it ASAP.

PR #271, opened 2021-11-04
Log exception set in IDiagnosticContext

angularsen:
Fixes: serilog/serilog-aspnetcore#270
Depends on: https://github.com/serilog/serilog-extensions-hosting/pull/56

### Changes
- `RequestLoggingMiddleware.LogCompletion` enriches with the exception set in `IDiagnosticContext`, unless an unhandled exception takes precedence.

### Testing
- Add tests on enriching with collected exception and favoring unhandled exception
- Functionally tested and Serilog request log event is correctly enriched with exception set in `IDiagnosticContext`

chrome_7VOs0WYLjb_masked

### DIFF
--- a/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
+++ b/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
@@ -86,7 +86,7 @@ namespace Serilog.AspNetCore
             // Enrich diagnostic context
             _enrichDiagnosticContext?.Invoke(_diagnosticContext, httpContext);
 
-            if (!collector.TryComplete(out var collectedProperties))
+            if (!collector.TryComplete(out var collectedProperties, out var collectedException))
                 collectedProperties = NoProperties;
 
             // Last-in (correctly) wins...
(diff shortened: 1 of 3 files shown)


angularsen:
Until https://github.com/serilog/serilog-extensions-hosting/pull/56 is merged and the dependency on `Serilog.Extensions.Hosting` is bumped, these tests will fail.

angularsen:
@nblumhardt Continuing from https://github.com/serilog/serilog-extensions-hosting/pull/56#issuecomment-1050368144:

> Since the stored exception is last in wins, should any exception caught in the request handling middleware take precedence over the exception stored in the context?

In our scenario, the response pipeline looks like this:
- MVC Action throws exception A
- Hellang middleware swallows it, calls `IDiagnosticContext.SetException()`, sets a problem details response
- Some other middleware throws exception B
- Serilog middleware goes wut do? ¯\\_(ツ)_\/¯ 

Going on my gut feeling, I would expect exception B to be logged since it was thrown last in the response pipeline. Also, I believe this requires no additional change to this PR.

What do you think?


I assume a new `serilog-extensions-hosting` nuget must be released before this PR can pass the tests on the new `SetExceptions()` method.

nblumhardt:
Sounds reasonable to me @angularsen 👍 

_Serilog.Extensions.Hosting_ 4.2.1-dev-00092 is now on NuGet and can be referenced from this repo to get the PR building 🚀

angularsen:
@nblumhardt Rebased and upgraded to prerelease nuget, tests are now green.

nblumhardt:
Looks good! Will be great to try this out in the wild :-)

angularsen:
Awesome! 🎉 I will give it a spin the coming week and report back.

PR #278, opened 2022-01-05
Update docs  for ASP.Net 6

darcythomas:


nblumhardt:
Thanks, Darcy! The `WebHost.UseSerilog()` call is not correct (should be `Host.UseSerilog()`) - I've written up an example https://blog.datalust.co/using-serilog-in-net-6/ and published a sample repo https://github.com/datalust/dotnet6-serilog-example - perhaps the README could draw on those?

Since .NET 6 is the current version, I think we should port/replace the current code samples in the README rather than add new sections (the older content will be obsolete soon, I guess).

Cheers!
Nick

darcythomas:
Oops my bad re webhost vs host.

.net 5 vs 6 maybe I can create a separate .net 5 page and link to that.

Since it there will be some people who are still gonna need that.

README.md
```diff
@@ -16,6 +16,77 @@ dotnet add package Serilog.AspNetCore
...
+
+app.UseRouting();
+
+app.UseAuthorization();
+
+app.MapRazorPages();
+
+
+
+// 3. Wrap app.run() in a try catch block
```
sungam3r, line 67:
```suggestion
// 3. Wrap app.Run() in a try catch block
```

README.md
```diff
@@ -16,6 +16,77 @@ dotnet add package Serilog.AspNetCore
 
 **Next**, in your application's _Program.cs_ file, configure Serilog first.  A `try`/`catch` block will ensure any configuration issues are appropriately logged:
 
+
+
+#### ASP.Net 6.0
+
+ASP.Net 6 combines the `Startup` and `Program.main`.
```
sungam3r, line 23:
```suggestion
ASP.Net 6 combines the `Startup` and `Program.Main`.
```

nblumhardt:
Thanks for the follow-up @darcythomas; I don't think we need an additional .NET 5 page - we don't currently have pages for .NET 4, Core 1 - 3, etc.. There's already plenty of info on .NET 5 setup out there, plus, anyone targeting .NET 5 now will likely have already started their project, and so already configured this.

darcythomas:
It might be more valuable to have a how to setup in ASP.Net MVC. 
Those who can will probably move to .Net 6 quite easily, but those stuck on older MVC (my current problem for one project) could do with instruction on how to get Serilog working on that platform.

**Shrug**

I guess moving forward is ultimately the better direction, so I'll delete the old setup mark down file.

README.md
```diff
@@ -16,59 +16,76 @@ dotnet add package Serilog.AspNetCore
...
 using Serilog.Events;
 
-public class Program
+var builder = WebApplication.CreateBuilder(args);
+
+//1. Add logger setup
+Log.Logger = new LoggerConfiguration()
+    .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
+    .Enrich.FromLogContext()
+    .WriteTo.Console()
```
darcythomas, line 33: This line is a bit problematic.

Printing to the console is slow. Like 100x slowdown in some of my tests.

In my code I have wrapped it in a `#IF DEBUG` block
I was wondering if the was a better (less ugly) way we should be disabling this in prod; and including that in the example given?

darcythomas:
@sungam3r Hey I think I have everything you noticed resolved. How do you feel about merging in to Main?

sungam3r:
I have no write access. I'm just a reviewer.

README.md
```diff
@@ -16,59 +16,76 @@ dotnet add package Serilog.AspNetCore
 
 **Next**, in your application's _Program.cs_ file, configure Serilog first.  A `try`/`catch` block will ensure any configuration issues are appropriately logged:
 
+_ASP.Net 6 combines the `Startup` and `Program.Main`._
+
+Update your `Program.cs` file like this
+
```
nblumhardt, line 22: These added lines are mostly redundant, let's leave them out.

README.md
```diff
@@ -91,7 +108,7 @@ The package includes middleware for smarter HTTP request logging. The default re
 
 As text, this has a format like:
 
-```
+``` bash
```
nblumhardt, line 111: Let's leave this plain; bash syntax highlights the "in", which is a bit odd.

README.md
```diff
@@ -16,59 +16,76 @@ dotnet add package Serilog.AspNetCore
...
+_ASP.Net 6 combines the `Startup` and `Program.Main`._
+
+Update your `Program.cs` file like this
+
 ```csharp
 using Serilog;
 using Serilog.Events;
 
-public class Program
+var builder = WebApplication.CreateBuilder(args);
```
nblumhardt, line 27: The placement of try/catch is important. Here's the example from datalust/dotnet6-serilog-example, cut down a little bit and merged with the existing sample.

```csharp
using Serilog;

Log.Logger = new LoggerConfiguration()
    .WriteTo.Console()
    .CreateLogger();

try
{
    Log.Information("Starting web application");

    var builder = WebApplication.CreateBuilder(args);

    builder.Host.UseSerilog(); // <-- Add this line
    
    var app = builder.Build();

    app.MapGet("/", () => "Hello World!");

    app.Run();
}
catch (Exception ex)
{
    Log.Fatal(ex, "Application terminated unexpectedly");
}
finally
{
    Log.CloseAndFlush();
}
```

Other features, like Serilog request logging, configuration, and the bootstrap logger, are covered below (this one is ultra-minimal, to emphasize the super-important exception handling block).

The original example is pretty opinionated, I think we should keep the same flavor as much as possible.

README.md
nblumhardt: Might be a good one to chat about in another thread :-)

nblumhardt (commented):
Thanks for all the updates @darcythomas! (And all the help, @sungam3r!)

I think we should keep the scope of this change minimal so that the structure of the README is still coherent, I've included some sample code in a comment above, let me know what you think.

README.md
darcythomas: I'll create an issue to discuss this in

darcythomas: #283

README.md
darcythomas: There is that better?

nblumhardt: Thank @darcythomas! I should have been more specific, I think we should keep the example to ☝️

darcythomas: @nblumhardt 
We can do. 
I chose to do a `dotnet new razor` and then add the minimal needed to get Serilog working correctly. 
Since I thought that most people who would be using serilog in a new project would busing that template.
(Although I may be projecting there haha)

So rather than only showing the minimum,  it would be better to show the smallest needed in a 'normal' or 'common' scenario.

Thoughts?
(If I haven't changed your mind, let me know and I can update things)

nblumhardt: Thanks for taking a look!

I think there's value in the first README example being compact/minimal, but we do also have a web project under `samples/Sample` that would be a good place to flesh out a full example with all the details (it could be recreated now using `dotnet new razor`) - I think the README might already link to it. What do you think?

cremor: A full sample (like https://github.com/datalust/dotnet6-serilog-example) would be very helpful.

glen-84:
As someone new to Serilog, it's really confusing and unfortunate having to find blog posts on Google with information about how to set this up. Why put it in a blog post instead of just adding it to the repository? .NET 6 is now over a year old, and .NET 7 has been released.

nblumhardt:
Looks like we stalled a bit here unfortunately, so I've just run through and made some minimal/quick updates to line things up with the .NET 6+ style of program/host initialization:

https://github.com/serilog/serilog-aspnetcore

Hope this helps. Thanks for kicking this one off @darcythomas 👍

Issue #279, opened 2022-01-06
Unable to publish .NET 6 console application due to package downgrade


PhilBroderickFS:
**Description**
When trying to publish a .NET 6 console application as a self-contained app, using the latest dev version of Serilog.AspNetCore, I receive errors due to package downgrades

![image](https://user-images.githubusercontent.com/76115748/148376187-abece71a-3238-478f-831c-89e92555cde5.png)

**Reproduction**
Add Serilog.AspNetCore package and try to publish a .NET 6 Console application, using `dotnet publish -c Release -r win-x64`

**Expected behavior**
Expected the application to be published with no errors.

**Relevant package, tooling and runtime versions**
Using Serilog.AspNetCore v4.1.1-dev-00241. 

**Additional context**
If I explicitly added Serilog.Settings.Configuration v3.3.0, I was able to successfully publish my console application. From the screenshot above, the package downgrade stems from an older version of Serilog.Settings.Configuration, so does this package need updated for the 'main' Serilog.AspNetCore package?

seriouz:
Yeah i can confirm this problem.

nblumhardt:
Fixed in the -dev package now.

PR #280, opened 2022-01-12
Bump Serilog.Settings.Configuration to latest version (3.3.0)

joaope:
We are trying to use some of the latest functionalities from `Serilog.Settings.Configuration`. It would be good to use the transitive dependency instead of forcing the latest manually.

I reckon this is a fairly safe PR. Let me know if I need to add something else.


nblumhardt:
Thanks! 👍 

Fixes #279

PR #265, opened 2021-09-23
Use IHttpRequestFeature.Path in RequestLoggingMiddleware to log RequestPath

jarronshih:
## Why

See #232 

`RequestPath` uses `RawTarget` which includes query data inside.

## What

`IHttpRequestFeature.Path` could be a better choose.

To not break the existing behavior, we make a option `RequestLoggingOptions.IncludeRawTargetPath` with default value `true`.

### Demo
Use sample.cs to test
```
# With options.IncludeRawTargetPath = false
2021-09-27 15:43:03.148 -07:00 [INF] Now listening on: https://localhost:5001
2021-09-27 15:43:03.181 -07:00 [INF] Now listening on: http://localhost:5000
2021-09-27 15:43:03.183 -07:00 [INF] Application started. Press Ctrl+C to shut down.
2021-09-27 15:43:03.185 -07:00 [INF] Hosting environment: Development
2021-09-27 15:43:03.187 -07:00 [INF] Content root path: C:\git\serilog-aspnetcore\samples\Sample
2021-09-27 15:43:03.942 -07:00 [INF] Hello, world!
2021-09-27 15:43:04.067 -07:00 [INF] HTTP GET / responded 200 in 183.3057 ms
2021-09-27 15:43:13.467 -07:00 [INF] HTTP GET /Index responded 404 in 0.4129 ms
2021-09-27 15:43:19.619 -07:00 [INF] HTTP GET /Index responded 404 in 0.0240 ms

# With options.IncludeRawTargetPath = true, which is default behavior
2021-09-27 15:43:35.651 -07:00 [INF] Now listening on: https://localhost:5001
2021-09-27 15:43:35.728 -07:00 [INF] Now listening on: http://localhost:5000
2021-09-27 15:43:35.745 -07:00 [INF] Application started. Press Ctrl+C to shut down.
2021-09-27 15:43:35.748 -07:00 [INF] Hosting environment: Development
2021-09-27 15:43:35.751 -07:00 [INF] Content root path: C:\git\serilog-aspnetcore\samples\Sample
2021-09-27 15:43:36.331 -07:00 [INF] Hello, world!
2021-09-27 15:43:36.496 -07:00 [INF] HTTP GET / responded 200 in 212.8998 ms
2021-09-27 15:43:42.311 -07:00 [INF] HTTP GET /Index?foo=bar responded 404 in 0.2009 ms
```

PR #286, opened 2022-02-02
Default to not including the query string in logged `RequestPath`

nblumhardt:
Follow-up from #265.

 * Default `IncludeQueryInRequestPath` to `false`
 * Bump major version (now 5.0.0)

As there's a major version bump I've taken the opportunity to mark the older `IWebHostBuilder` extensions in this package as `Obsolete`.


nblumhardt:
👋 

I'll send this through to dev tomorrow, to keep the ball rolling :-)

PR #317, opened 2023-01-23
Microsoft.Extensions.Logging set to min version 2.2.0 across

AlthalusDGr8:


nblumhardt:
Thanks for sending this. We've switched the package to a new versioning strategy that will track the underlying versions of MEL etc. that are in use (see #326), so we won't be able to merge this PR. Thank you again, though!

PR #338, opened 2023-06-29
Remove obsolete APIs and little cleanup

sungam3r:


nblumhardt:
Thanks @sungam3r :+1:

I'm on vacation but should be back to full capacity next week.

SimonCropp:
@nblumhardt LGTM. do we bump a major for this change?

nblumhardt:
@SimonCropp this package's versioning tracks ASP.NET Core's, so I think this falls in naturally with the 8.0 release (which we'll need to start assembling soon).

Bumping major to 8.0.0 on `dev` seems fine, as long as the ASP.NET Core package dependencies also move forward to 8 before we release it 👍

Issue #345, opened 2023-09-18
Add EnrichDiagnosticContextAsync as async addition


hbunjes:
**Is your feature request related to a problem? Please describe.**
I want to enrich the http logging with Serilog by parts of the request body. Therefore, I read from the request body which must be done async. However, if a set EnrichDiagnosticContext to an async method like the following I get the VSTHRD101 warning:
```
options.EnrichDiagnosticContext = async (context, httpContext) =>
    {
        await LogHelper.EnrichFromRequestAsync(context, httpContext);
    };
```

**Describe the solution you'd like**
In addition to the `Action<IDiagnosticContext, HttpContext>? EnrichDiagnosticContext`, there should be a `Func<IDiagnosticContext, HttpContext, Task>? EnrichDiagnosticContextAsync` that can be called asynchronously and returns just a Task.

hbunjes:
I've added a PR to resolve this.

PR #346, opened 2023-09-20
Add EnrichDiagnosticContextAsync to allow async calls

hbunjes:
https://github.com/serilog/serilog-aspnetcore/issues/345

If you want to add properties to the diagnostic context that require async calls (e.g. reading request body), EnrichDiagnosticContextAsync allows to hand over your own async method.

### DIFF
--- a/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
+++ b/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
@@ -28,6 +28,7 @@ class RequestLoggingMiddleware
     readonly DiagnosticContext _diagnosticContext;
     readonly MessageTemplate _messageTemplate;
     readonly Action<IDiagnosticContext, HttpContext>? _enrichDiagnosticContext;
+    readonly Func<IDiagnosticContext, HttpContext, Task>? _enrichDiagnosticContextAsync;
     readonly Func<HttpContext, double, Exception?, LogEventLevel> _getLevel;
     readonly Func<HttpContext, string, double, int, IEnumerable<LogEventProperty>> _getMessageTemplateProperties;
     readonly ILogger? _logger;
@@ -42,6 +43,7 @@ class RequestLoggingMiddleware
 
         _getLevel = options.GetLevel;
         _enrichDiagnosticContext = options.EnrichDiagnosticContext;
+        _enrichDiagnosticContextAsync = options.EnrichDiagnosticContextAsync;
         _messageTemplate = new MessageTemplateParser().Parse(options.MessageTemplate);
         _logger = options.Logger?.ForContext<RequestLoggingMiddleware>();
         _includeQueryInRequestPath = options.IncludeQueryInRequestPath;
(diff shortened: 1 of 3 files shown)


nblumhardt:
Thanks for sending this!

I think this feature would be find if it cleanly slotted in, but as you found in the implementation, it's not completely equivalent to the existing `EnrichDiagnosticContext` option 🤔  - not sure how much that matters, but does suggest people might observe some unexpected differences in collected data when switching between the synchronous and asynchronous overloads.

The `EnrichDiagnosticContext` option we have right now also avoids adding overhead when the request completion event's level is not enabled, which is harder to do for the async version.

Just zooming out, the alternative implementation of this scenario might look like:

```csharp
class RequestBodyEnrichmentMiddleware(RequestDelegate next, IDiagnosticContext diagnosticContext)
{
    public async Task InvokeAsync(HttpContext context)
    {
        await EnrichFromRequestAsync(context.Request);
        await next(context);
    }

    async Task EnrichFromRequestAsync(HttpRequest request)
    {
        // diagnosticContext.Set(...)
    }
}
```

Perhaps giving that example in the README would go far enough?

hbunjes:
Nicholas, thank you for your valuable feedback. I totally get your point. 

The reason I think this is still a good place to integrate is that you already have some good features like "GetLevel" in here and it's just a good match to have the logging information at one point.

I've changed the implementation to behave like the sync call. It's always called in finally block (and makes sure it doesn't create new Exceptions, just like the second call of "LogCompletion" in the exception filter clause). It's also taking the information if the method should be called based on GetLevel and current log level.

I'll check why two tests fail (it seems I've introduced a bug between my last tests and commiting the solution) and update the PR then.

src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
```diff
@@ -54,31 +56,58 @@ public async Task Invoke(HttpContext httpContext)
...
+            // The elapsedMs is required for calculating the level and for LogCompletion.
+            // The level is required in the finally block.
+            when ((elapsedMs = GetElapsedMilliseconds(start, Stopwatch.GetTimestamp())) >= 0
+                  && (level = _getLevel(httpContext, elapsedMs, ex)) >= 0
+                  && LogCompletion(httpContext, collector, 500, elapsedMs, logger, level, ex))
         {
         }
         finally
         {
+            await CallEnrichDiagnosticContextAsync(httpContext, logger, level);
```
nblumhardt, line 85: I may be off track, but won't this be called _after_ the `LogCompletion` calls in the `try` and `catch/when` blocks?

sungam3r:
I would prefer to moce from `public Action<IDiagnosticContext, HttpContext>? EnrichDiagnosticContext { get; set; }` to `public Func<IDiagnosticContext, HttpContext, Task>? EnrichDiagnosticContext { get; set; }`

Gonkers:
Is there any movement expected on this?

Dong-Ruipeng:
Is there any movement expected on this?

PR #347, opened 2023-09-25
use Stopwatch for ElapsedMilliseconds

SimonCropp:


src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
```diff
@@ -53,20 +53,19 @@ public async Task Invoke(HttpContext httpContext)
     {
         if (httpContext == null) throw new ArgumentNullException(nameof(httpContext));
 
-        var start = Stopwatch.GetTimestamp();
+        var stopwatch = Stopwatch.StartNew();
```
bartelink, line 56: While this is less cryptic for people scanning (and I also agree that the math is v error prone wrt overflow and/or getting int64 vs double conversions wrong), this line does allocate so I went the other way in my helpers in https://github.com/jet/propulsion/blob/master/src/Propulsion/Internal.fs#L9-L18
Having said that, I also do have lots of places where I actually want/need it to be in ticks so I can add lots together and then render variously as seconds, or ms.
I'd also note that the jitter and the compiler also have plenty special casing to make the math and/or inlining be more efficient than you'd expect for the code as it was
Also .net 7 adds a `System.Diagnostics.Stopwatch.GetElapsedTime()` so maybe that can be used?

nblumhardt: :+1: the shift to `GetTimestamp()` was to avoid a wasted allocation here; using `GetElapsedTime()` would be the best-of-both-worlds solution 👍

PR #349, opened 2023-10-11
Collect trace and span ids in request completion events

nblumhardt:
Reported in: https://github.com/serilog/serilog-aspnetcore/issues/207#issuecomment-1756844594

The dependencies and their versions here are temporary, we'll soon be updating this package to .NET 8 and will ideally RTM the upstream pieces then, too.

### DIFF
--- a/samples/Sample/Program.cs
+++ b/samples/Sample/Program.cs
@@ -1,4 +1,5 @@
 using Serilog;
+using Serilog.Templates;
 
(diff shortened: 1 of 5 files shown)


nblumhardt:
This is part of a wider set of related changes that enable trace and span ids, so I'll push this through, but reviews after-the-fact are welcome.

PR #364, opened 2024-03-05
Fix properties example to use LogContext

### LINKED ISSUE #363: Small typo with properties example in documentation
Believe there is a bug in the example in https://github.com/serilog/serilog-aspnetcore#pushing-properties-to-the-iloggert. It should be **LogContext.PushProperty** rather than **logger.PushProperty**. Alternatively or additionally it can also point to https://github.com/serilog/serilog/wiki/Enrichment#the-logcontext for more information.

However it would be be a nice to have **ILogger.PushProperty** which then forwards to **LogContext.PushProperty**.

RandomBuffer:
Fixes #363

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,12 +276,12 @@ using (logger.BeginScope(new Dictionary<string, object>
 }
 ```
 
-The code above results in the same outcome as if you would push properties in the **ILogger** in Serilog.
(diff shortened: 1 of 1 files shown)


nblumhardt:
Thanks!

Issue #389, opened 2024-11-28
Add elapsedMs parameter to the EnrichDiagnosticContext delegate


jmenziessmith:
**Is your feature request related to a problem? Please describe.**
I want to use the diagnostic context to log a structured object containing enhanced request information. I would like this to include the elapsed duration, but the value is not accessible outside of the middleware. I believe this is a simple fix, to just include the elapsedMs value as a parameter to the EnrichDiagnosticContext action.

**Describe the solution you'd like**

[RequestLoggingMiddleware.cs#L30](https://github.com/serilog/serilog-aspnetcore/blob/dev/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs#L30)
```csharp
readonly Action<IDiagnosticContext, HttpContext>? _enrichDiagnosticContext; // current
readonly Action<IDiagnosticContext, HttpContext, double>? _enrichDiagnosticContext; // proposed
```

[RequestLoggingMiddleware.cs#L85](https://github.com/serilog/serilog-aspnetcore/blob/dev/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs#L85)
```
_enrichDiagnosticContext?.Invoke(_diagnosticContext, httpContext); // current
_enrichDiagnosticContext?.Invoke(_diagnosticContext, httpContext, elapsedMs); // proposed
```

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

nblumhardt:
Thanks for dropping us a line!

How does the duration you want to pass through the `IDiagnosticContext` differ from the one computed by the middleware?

jmenziessmith:
Hi. Thanks for the quick reply.

I don't expect the duration to differ - I would expect the same value. The desire is to be able to access that value when enriching the DiagnoticContext.

nblumhardt:
Ah sorry, I see I misunderstood 😅 - thanks for the clarification.

jmenziessmith:
@nblumhardt I have raised a PR with an alternative solution in order to avoid introducing breaking changes to the EnrichDiagnosticContext action. 

https://github.com/serilog/serilog-aspnetcore/pull/391

PR #391, opened 2025-01-06
Add option to include ElapsedMs in the HttpContext prior to invoking EnrichDiagnosticContext

jmenziessmith:
#389 

Adds `AddElapsedToHttpContext` to _RequestLoggingOptions_. When set to true, will add the _elapsedMs_ value to the HttpContext Items collection prior to invoking _EnrichDiagnosticContext_

I'm suggesting this approach in order to avoid any breaking changes to _EnrichDiagnosticContext_

### DIFF
--- a/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
+++ b/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
@@ -32,6 +32,7 @@ class RequestLoggingMiddleware
     readonly Func<HttpContext, string, double, int, IEnumerable<LogEventProperty>> _getMessageTemplateProperties;
     readonly ILogger? _logger;
     readonly bool _includeQueryInRequestPath;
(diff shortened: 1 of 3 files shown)


nblumhardt:
Thanks for the PR, and sorry about the slow reply @jmenziessmith - took some time off over the Christmas/New Year period :-)

I think this solution might be a good workaround if you're able to add a `Stopwatch` instance to `HttpContext.Items` via custom middleware. I think API discoverability is enough of an issue that it probably isn't the ideal approach for the Serilog middleware to implement directly.

I can see it would have been nice to consider this requirement when designing the Serilog middleware originally; unfortunately, it might be too late to add it in a discoverable-but-nonbreaking manner.

Assuming `Activity.Current` is non-null in the enricher callback, the elapsed time of the activity would also indirectly reveal the elapsed time of the web request - this could be a good middle-ground to explore, let me know if you need more info!

PR #404, opened 2026-08-15
Suppress exceptions for aborted requests

### LINKED ISSUE #372: Do not log exception call stacks for cancellations in RequestLoggingMiddleware
**Is your feature request related to a problem? Please describe.**
When a client cancels a request the RequestLoggingMiddleware will log the stack trace of the `TaskCanceledException`.

**Describe the solution you'd like**
`RequestLoggingMiddleware` should not log exception strack traces for cancellations.

**Describe alternatives you've considered**
Maybe make it configurable. But I can't see a reason why anyone would want stack traces for cancellations.

**Additional context**
Related to #341 but not the same. You might want to log "real" exceptions but cancellation exceptions are always spam.
This could be implemented in the same way as [DeveloperExceptionPageMiddlewareImpl](https://github.com/dotnet/aspnetcore/blob/0e7349e41a112a0c62f364ce9666a1a7dd4dc2fd/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddlewareImpl.cs#L115) or [ExceptionHandlerMiddlewareImpl](https://github.com/dotnet/aspnetcore/blob/0e7349e41a112a0c62f364ce9666a1a7dd4dc2fd/src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddlewareImpl.cs#L117) implements it.
It would be a good idea to also fix the status code for cancellations at the same time (related to #344).

Solution-lab:
Fixes #372.

## Summary

`RequestLoggingMiddleware` no longer attaches an escaping `OperationCanceledException` or `IOException` to the completion event when `HttpContext.RequestAborted` is cancelled.

The completion event is still emitted, the exception continues to propagate through the middleware pipeline, and exceptions collected through `IDiagnosticContext.SetException()` remain attached.

## Rationale

This follows ASP.NET Core's request-abort handling. `IOException` was included intentionally ([aspnetcore#46330 discussion](https://github.com/dotnet/aspnetcore/pull/46330#discussion_r1099323588)), and the exception is omitted to avoid noisy, non-actionable stack traces ([Tratcher's comment](https://github.com/dotnet/aspnetcore/pull/46330#discussion_r1093576244)).

## Tests

Tests cover both cancellation exception types, non-aborted cancellation, and preservation of collected exceptions.

### DIFF
--- a/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
+++ b/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs
@@ -94,10 +94,16 @@ class RequestLoggingMiddleware
             (activity.TraceId, activity.SpanId) :
             (default(ActivityTraceId), default(ActivitySpanId));
 
+        var isRequestAborted =
+            (ex is OperationCanceledException || ex is IOException) &&
+            httpContext.RequestAborted.IsCancellationRequested;
+
+        var eventException = isRequestAborted ? collectedException : ex ?? collectedException;
+
         var evt = new LogEvent(
             DateTimeOffset.Now,
             level,
-            ex ?? collectedException,
+            eventException,
             _messageTemplate,
             properties,
             traceId,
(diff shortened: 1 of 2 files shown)


nblumhardt:
Thanks! Current trajectory will have this out in a v11.0 in a couple of months.

